PR #65, opened 2014-12-16
Add epetra matrix format, with aztecoo and amesos, better integration

wortiz:
Adds the epetra matrix format natively to goma, only supported
with AztecOO and Amesos solver

This attempt has better integration with the various methods that touch the matrix directly outside of load lec.

Passes the test suite.

Also passes the test suite with minimal differences on all problems using epetra instead of msr or vbr. Problems that cannot be run with native epetra fail with EH.

Note: There are some misbehaving problems though (weld, overset_ball, porous_deform) that also produce different results with the msr to epetra translation, and the native epetra results match.

New syntax for input file documented in `docs/epetra_matrix_format.md`


acochrane:
Because we changed the history since this pull request was issued, it could not be automatically merged. In order to bring these changes into the master branch, we cherry-picked the commit from wortiz' `epetra_full_pull' branch.

PR #68, opened 2015-02-09
Added Galerkin least-squares type stabilization using the continuity operator

dshari:
Hi all,

This adds a stabilization method using the continuity operator.  It should help in high Reynolds number situations.

I ran it by the test suite, and I got some of those funny .exoII failures.  I wouldn't mind if somebody else ran it by the test suite to independently confirm it is good to go.

Daniel

### DIFF
--- a/include/rf_solver.h
+++ b/include/rf_solver.h
@@ -151,6 +151,7 @@ int Gfill_zeros;        /* number of nonzeros in fill matrix for the
 int PSPG;               /* 1 means pressure stabilized Petrov-Galerkin is used */
 int PSPP;               /* 1 means pressure stabilized polynomial projection is used */
 double PS_scaling;      /* This term is a constant scaling for the PSPG or PSPP term */
+int Cont_GLS;          /* 1 means continuity stabilization is used */
 
 int Filter_Species, filter_species_material_number;
 double c_min, c_max;
--- a/src/mm_fill_terms.c
+++ b/src/mm_fill_terms.c
@@ -156,6 +156,7 @@ extern FSUB_TYPE dsyev_(char *JOBZ, char *UPLO, int *N, double *A, int *LDA,
 *  heat_source                 double
 *  ls_modulate_heatsource 
 *  calc_pspg
+*  calc_cont_gls
 *  assemble_ls_latent_heat_source
 *  assemble_acoustic		int 
 *  assemble_acoustic_reynolds_stress int
@@ -2452,6 +2453,12 @@ assemble_momentum(dbl time,       /* current time */
 					   of the vertices for Q1. It comes from
 					   the routine "element_velocity." */
 
+  //Continuity stabilization
+  dbl continuity_stabilization;
+  dbl cont_gls;
+  CONT_GLS_DEPENDENCE_STRUCT d_cont_gls_struct;
+  CONT_GLS_DEPENDENCE_STRUCT *d_cont_gls = &d_cont_gls_struct;
+
   int *n_dof = NULL;
   int dof_map[MDE];
   
@@ -2711,6 +2718,12 @@ assemble_momentum(dbl time,       /* current time */
 
   (void) momentum_source_term(f, df, time);
 
+  //Call continuity stabilization if desired
+  if(Cont_GLS)
+    {
+      calc_cont_gls(&cont_gls, d_cont_gls, time, pg_data);
+    }
+
   /*
    * Residuals_________________________________________________________________
    */
@@ -2906,6 +2919,16 @@ assemble_momentum(dbl time,       /* current time */
 		     * source term. */
 		    source = element_particle_info[ei->ielem].source_term[i];
 		}
+	      
+	      //Continuity residual
+	      continuity_stabilization = 0.0;
+              if (Cont_GLS)
+		{
+		  for (p=0; p<VIM; p++){
+		    continuity_stabilization += grad_phi_i_e_a[p][p];
+		    }
+                  continuity_stabilization *= cont_gls *d_area;
+	        }
 
 	      /*
 	       * Add contributions to this residual (globally into Resid, and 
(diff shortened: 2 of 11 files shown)


rrrao:
Does it default to off? Weird about exo diffs.

dshari:
Yes, the default is off.  

Weston mentioned something about the parallel version of guts, which I used, producing slightly different simulation results.  GUTS then sees a different .exoII file and reports a failure, even though the differences are small.

Daniel

wortiz:
Check the differences, I'm guessing you built with sparse 1.4 which produces quite small differences, but @acochrane was having issues with guts as well so make sure nothing is breaking.

Kris's machine has sparse 1.3 built with gcc 4.8, I currently don't have access to the machines so I haven't run guts yet on the latest master change to see if there are any issues.

The parallel guts fixes the norm compares but other than that 1 mysterious run you had I haven't been able to produce different exodus results with pguts.

It might be a good idea to rebase guts to sparse 1.4 in the future.

dshari:
I ran the test suite on the master and compared its results with my branch's results.  It looks like most of the problems I was referring to also appear on the current master, and most of the differences are related to convergence rates.  I didn't get a chance to try Kris' machine yet to see if the differences are due to library/compiler changes.

There are 11 problems that fail the .exoII test for the current master: foam, psat_rdl, parallel_x, unsat_infiltrate, sat, psat_ant, pipe.lu.q2p1, MCD, ldc_cont-g1, ld_cont-g0, and current.  We should look into rebasing guts some time soon.

My changes add two new .exoII failures: dea and p_dea_pspg.

The dea problem shows a slightly different convergence rate for the last iteration.

The p_dea_pspg problem has one node with a different div(v) value and a separate node with a different p value.  The values are O(10^-10) and O(10^-7), with relative differences O(10^-2) and O(10^-3).  I don't think there is a need to worry.  p_dea_pspg is the parallel version of another problem, dea_pspg, which doesn't show any errors.  I may have exposed a small parallel issue though.

include/rf_solver.h
```diff
@@ -151,6 +151,7 @@ int Gfill_zeros;        /* number of nonzeros in fill matrix for the
 int PSPG;               /* 1 means pressure stabilized Petrov-Galerkin is used */
 int PSPP;               /* 1 means pressure stabilized polynomial projection is used */
 double PS_scaling;      /* This term is a constant scaling for the PSPG or PSPP term */
+int Cont_GLS;          /* 1 means continuity stabilization is used */
```
saroberts: Is ```Cont_GLS``` initialized and/or set appropriately before use in all cases?

saroberts:
I briefly looked through your code, and generally it looks good.  However, I think it would be good for somebody more familiar with GLS (Rekha) to review the code thoroughly.

With regards to the test diffs, my first reaction is that they are no big deal and you could rebase.  However, it is a bit concerning that adding a section of code should NOT affect problems that don't affect that code.  Suggests that there may be memory leak issues, uninitialized values, or parallel consistency issues.  We know these things exist in GOMA and we should do our best to track that down.  However, I have no problem pulling this change after a review.

include/rf_solver.h
dshari: Cont_GLS is initialized in src/mm_input.c.  I followed the implementation of PSPG.

rrrao:
Daniel,

Maybe we can look at the diffs on Friday and re-bless the test suite?

Rekha

dshari:
Yes, let's look at everything on Friday.  I agree with Scott, we should try to track down the new issues from my code and fix them while we have a chance.

I have a feeling similar issues will arise with the VE GLS stuff, so this way we can be better prepared for those future problems.

wortiz:
I ran the test suite on master and on your branch on Kris's machine.

foam and p_level_set_porous differed because I didn't check with the gcc-4.8 branch, we should probably rebaseline on gcc-4.8 for guts master. But no other changes occurred for master.

For your branch however two additional problems came up with issues:

* `dea` had a difference in convergence rate which would probably be okay to rebaseline but if your new stuff is off by default the old behavior should have not changed

* `p_dea_pspg` Has some smallish differences in P and DIVV

The rest of the differences that you are seeing I think are from sparse 1.4 instead of sparse 1.3, I believe we switched to sparse 1.4 for the build script and build notes for public versions as sparse 1.3 from netlib needs to be modified completely to run on modern compilers.

prschun:
Lets sit down this Friday and figure this out.  I have some ideas here.   Scott's concern is also my concern.    Thanks!

dshari:
Hi all,

I updated the commit based on comments from Rekha.  There was also a small indexing bug I had to fix as well.  

I did a couple of Jacobian checker tests, and it passed with no issue.  I also ran it by the test suite again, and now its results are the same as the current master with my machine/libraries.  I think this is ready for prime time, but I would also welcome more feedback.

Daniel

src/mm_fill_terms.c
```diff
@@ -29022,6 +29097,219 @@ return 0;
...
+	    {
+	      for(j=0; j<ei->dof[var]; j++)
+		{
+		  if(Cont_GLS==1 || h_elem==0)
+		    {
+		      d_tau_dmesh[b][j] = 0.;
+		    }
+		  else if(Re > 1.0)
+		    {
+		      d_tau_dmesh[b][j] = rho/mu_avg *U *U *pg_data->hh[b][b] *pg_data->dh_dxnode[b][j];
```
saroberts: Just a minor style comment.  I would much prefer either
`mu_avg*U*U*pg`
or
`mu_avg * U * U * pg`
rather than 
`mu_avg *U *U *pg`
simply because the latter looks like you're dereferencing a pointer rather than multiplying.  No big deal, though.

rrrao:
Daniel,
It looks  like you fixed all the things we discussed last w

PR #73, opened 2015-04-30
April end LIGHT_JUMP

rbsecor:
Added the routines for LIGHT(P/M)_JUMP_BCs (discontinuous light intensity @ interfaces) ; modified CAP_REPULSE_TABLE to be more physical (lubrication squeezing pressure plus drag from gas escape).

Test suite results are the same as April 28 push.


dshari:
The test suite also passes on my machine

PR #76, opened 2015-05-12
Initialize WeEe array in assemble_PPPS_generalized

wortiz:
Don't merge without examining assemble_PPPS_generalized.

This initializes the WeEe array to 0, some values from WeEe array that were not initialized or set were making it into lec->J.

This initialization may need to be in the `for (ipass = 0; ipass < npass; ipass++)` loop. Someone with knowledge of this assembly should look and see if that is the right option.

### DIFF
--- a/src/mm_fill_terms.c
+++ b/src/mm_fill_terms.c
@@ -18388,8 +18388,6 @@ assemble_PPPS_generalized(Exo_DB *exo)
   double scaling_over_visc_e, my_multiplier=1.0;
   double R_new[MDE], R_old[MDE];  /*old and new resid vect pieces for the pressure equation */
 
-
-  
   asdv(&A, ei->ielem_dim * ei->ielem_dim);
 
   npass = 1;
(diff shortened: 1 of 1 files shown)


acochrane:
I have no knowledge of the method or its implementation. But looking at where WeEe was used, would it be fine to arrange the memset to right before it's algorithmic assignment, similar to the other arrays whose pointers are allocated in the same line? (i.e.  De_hat, S, We).

I'm looking at the loop beginning at line 18680, and the placement of the memset as such:
memset(WeEe    , 0, sizeof(double)*(DIM+1)*(MDE));
for(i=0; i < ei->ielem_dim+1; i++) {
  for(j=0; j < ei->dof[eqn] ; j++) {
    sum = 0.;
    for(k=0; k < ei->ielem_dim + 1; k++) {
	sum += We[i][k]* Ee_hat  [k][j];
    }
    WeEe[i][j] = sum; 
  }
}

wortiz:
Memset moved to suggested location.

PR #78, opened 2015-05-18
Use trilinos export makefile rather than manually specifying libraries

wortiz:
Allows Trilinos 10-12 to be used without setting libraries manually and instead uses the Makefile export that cmake creates

CXXSTD can be set in settings.mk to use trilinos 12 which requires c++11:

    CXXSTD = -std=c++11

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -176,25 +176,13 @@ SEACAS_LIB ?= -L$(SEACAS_TOP)/lib  $(EXODUSII_LIB)  $(NETCDF_LIB) \
 # Trilinos install directory (include, lib)
 TRILINOS_TOP ?= $(GOMA_LIBS)/trilinos-11.8.1-Built
 
-# KOMPLEX library (not included by default)
-KOMPLEX_LIB ?=
 
-# AZTEC library
-# Aztecoo is required.
-AZTEC_LIB ?=  -laztecoo -lifpack 
-
-# AMESOS library
-
-AMESOS_LIB ?= -lamesos -lepetraext -lepetra 
-
-# TEUCHOS library
-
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -176,25 +176,13 @@ SEACAS_LIB ?= -L$(SEACAS_TOP)/lib  $(EXODUSII_LIB)  $(NETCDF_LIB) \
...
-AZTEC_LIB ?=  -laztecoo -lifpack 
-
-# AMESOS library
-
-AMESOS_LIB ?= -lamesos -lepetraext -lepetra 
-
-# TEUCHOS library
-
-TEUCHOS_LIB ?= -lteuchoscomm -lteuchoscore -lteuchosnumerics -lteuchosparameterlist -lteuchosremainder
+include $(TRILINOS_TOP)/include/Makefile.export.Trilinos
```
saroberts: Nice find!

src/Makefile
```diff
@@ -315,7 +303,8 @@ FFLAGS = $(CCFLAGS) $(F_WARN_FLAGS) $(DEFINES) $(INCLUDES)
 CFLAGS = $(CCFLAGS) $(C_WARN_FLAGS) $(DEFINES) $(INCLUDES)
 
 # C++ compiler flags
-CXXFLAGS = $(CCFLAGS) $(C_WARN_FLAGS) $(DEFINES) $(INCLUDES) -Wno-deprecated
+CXXSTD ?=
```
saroberts: You might consider turning this on by default and testing its behavior.  Much of the Trilinos development nowadays uses C++11, and I think in their next release they will require it.  Better get used to it now.

PR #80, opened 2015-06-12
SuiteSparse and Trilinos' SEACAS

acochrane:
The UMFpack standalone dissapeared from UFL hosting site.
SuiteSparse is now the available UMFpack source.
Also SEACAS archive is no longer supported on Sourceforge.
The Trilinos build process in build-goma-depencies.sh has been
modified to build the SEACAS that it ships with.

The src/Makefile has also been updated to reflect these changes

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,33 @@
+# Goma Dependincies build script
+
+Name
+
+	build-goma-dependencies.sh
+
+Synopsis
+
+	build-goma-dependincies.sh -jN [library install path]
+
+Description
+
+	This script attemps to download, compile and organize all the dependencies of goma.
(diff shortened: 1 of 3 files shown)


wortiz:
Issues I've had with this script:

Trilinos could not finish linking because libcurl doesn't exist on my system

HDF5 is expected to be on LD_LIBRARY_PATH (shared library)

The Trilinos_EXTRA_LD_FLAGS needs to be at the end of the linking in the Makefile or else it can't find the hdf5 libraries.

wortiz:
Test suite passes.

Script works on ubuntu 14.04, centos 6, and centos 7.

acochrane:
Incorporated Weston's changes, and disabled curl requirement of netcdf with --disable-dap

PR #81, opened 2015-06-16
Add example settings.mk for old style goma lib layout

wortiz:
Add an example of how to format settings.mk for when seacas is installed separately from trilinos.


wortiz:
I'm going to merge this as the settings file works with minor adjustments on a few builds with the old separate seacas.

PR #84, opened 2015-07-21
Add advanced capabilities pdf to docs/

wortiz:


saroberts:
I still think at some point we should think about converting the manuals to LaTeX and keeping the source in repository.  Just a thought.

wortiz:
I think that was one of the better ideas, we only had an old version of framemaker here that didn't have good export support so it would've been a lot of hand conversion. Maybe we can see if newer versions of framemaker have better export ability now that Randy is here.

saroberts:
Agreed.  As long as you can get the text out, you should be able to write some sort of python script to add in the necessary LaTeX structure.  That and a custom cls file and you should be able to make it look however you want.  Might have to do some manual work on equations and images, but would be worth it.

PR #85, opened 2015-08-14
Change the trilinos build scripts to build 12.2.1

wortiz:
Trilinos 12.2.1 no longer requires a patch to blot/fastq

No changes in test suite with this version of trilinos


acochrane:
Builds and runs knife coater example in Slackware 14.0

PR #86, opened 2015-08-18
Add extra check in mm_fill to only access a variable if it exists

wortiz:
From the problem Louis had, if a problem was broken in such a way that one processor had no nodesets then this would segfault as Proc_NS_List was not allocated.

Passed test suite.


prschun:
I knew this.    This is an oldy.     Sorry I couldn't head it off at the pass.
Randy

PR #92, opened 2016-02-12
Update makefile to cause a full recompile when a header file changes and when makefiles or settings.mk are changed

wortiz:
Update makefile to cause a full recompile when a header file changes and when makefiles or settings.mk are changed.

Fixes the issue where if you changed a header file or Makefile you had to remember to `make realclean`

Did not know about $< symbol for makefiles before.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -692,8 +692,8 @@ PREFIX_INC = $(PREFIX)/include
 # ============
 
 # Object files depend on includes and the makefile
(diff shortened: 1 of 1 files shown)


acochrane:
This will at the very least eliminate the necessity of running 'make clean' between modifying a header file and running 'make'.
I think that's as good as it gets without setting up dependencies for each individual object file, or using compiler dependent auto-dependency functionality.

PR #93, opened 2016-02-13
Add stratimikos support with file specification

acochrane:
Build script "build-goma-dep-trilinos-12.sh" now compiles trilinos with stratimikos support

settings.mk-example has been updated with pieces for stratimikos support

Stratimikos support requires C++11
https://en.wikipedia.org/wiki/C%2B%2B11


acochrane:
The noc++11 build script no longer tries to build with stratimikos and has been updated for new blas site layout.

Both build scripts produce suitable gomalib directories for use with goma, enabling stratimikos with c++11 with and disabling stratimikos without c++11.

acochrane:
Should we revise to add instructions for the build-trilinos-stratimikos script?

acochrane:
Improved instructions for using stratimikos

wortiz:
Passes test suite so blas update didn't break anything.

Stratimikos works as a solver.

Thanks for updating the docs and the scripts.

PR #96, opened 2016-02-23
Fix bug on scale vector calculations used in arc length continuation.…

tjiptowi:
I commented the offending lines. I am not sure what it is supposed to be doing. Arc length continuation is working now for P1 basis function.


dshari:
It passes the test suite.

PR #104, opened 2016-03-14
Add SHELL_VOLUME into Post Processing Volumetric Integration capability

tjiptowi:


wortiz:
From a code cleanliness standpoint you might want to add curly braces to the switch statement (similar to the one below it) so your new variables stay local to that case.

wortiz:
Code passed test suite, and curly braces were added by Kris.

PR #106, opened 2016-03-24
Protect the user from running into nans when using shells BCs that re…

dshari:
…ly on fv->snormal


dshari:
I updated this pull request to include a fix for the lubrication velocity field in post processing when the SHELL_FILM equations are in use.

PR #107, opened 2016-04-04
Removing faulty code for the FILM_H and FILM_PARTC equations

dshari:
This fixes a segfault for my gravure problem.  It looks like it was a quick and dirty fix for another problem that is no longer needed.


wortiz:
Passes test suite.

PR #108, opened 2016-05-12
Update Trilinos, SuperLU_DIST, MUMPS in build script

wortiz:
Updates build script to trilinos 12.6.3 which supports superlu dist 5.0.0 and mumps 5.0.1

Works on ubuntu and passes test suite.

Need someone else to test.


wortiz:
Fixed issue caused by not removing old mumps checksum causing suitesparse and matio to say they had invalid checksums.

wortiz:
Fixed issue where cmake was used before it was on the path

dshari:
Looks like everything worked after the cmake fix.

PR #109, opened 2016-06-06
Print out err instead of iteration number when stratimikos did not converge

wortiz:
Similar to ill and other such things for aztec, we were not seeing an error when using stratimikos but just a low iteration number.

This adds a check to make sure to only print iterations if stratimikos has converged otherwise it prints err to indicate some error occurred.


acochrane:
.buildgoma

wortiz:
Closing to create  new request for jenkins.

PR #111, opened 2016-06-16
Gomathon June 2016 update

rbsecor:
Added U/V/W_PARABOLA_BC for 2D inlet flow profiles
Updated PHOTO_CURING model
Passes test suite in same fashion as last update


wortiz:
Passed test suite.

PR #114, opened 2016-06-22
June2016 userbc b

rbsecor:
Converted a couple user routines to standard BCs for increased
usability.  U/V/W_PARABOLIC provides parabolic and power-law
inflow velocity profiles for 2D problems. FILLET_BC is a geometry
routine on R_MESH_NORMAL for radiused corners.

guts results are the same as previous.


wortiz:
Passed test suite, Robert also reported it fixed his CA issue.

PR #117, opened 2016-07-05
Replaced the static makefiles (with settings.mk) with cmake

natis1:
Cmake allows you to automatically generate and configure makefiles. It is much more versatile then traditional makefiles. It requires the cmake package, which, being a GNU program, should be on every Linux distribution.

If you are an average user and need a build quickly, just run "cmake ." and then "make" or "make install." But, and here is the advantage of Cmake, not only does it obsolete manual settings files (like settings.mk), and manual makefiles, but it also lets you configure settings much quicker. Now you just have to run cmake . and then ccmake . (or cmake-gui . ) and reconfigure anything. Did it not find the libraries automatically? give it the correct location. Do you want to compile goma with the debug hooks? enable debug mode (this also replaces make debug). Do you want some special compiler flags or definitions? Write whatever flags you want. Not enough control over CMake? Edit the CMakeLists.txt file to whatever you want.

Try it out by downloading my head fork.

Be sure not to commit any of the autogenerated cmake files or it might not work on other systems.


wortiz:
A few things to start off with:

This is going to need an update to the readme as it changes what everyone is used to and preferably an example cmake configure for debugging and for regular builds

An example of what a cmake configure script might look like from the seacas repo https://github.com/gsjaardema/seacas/blob/master/cmake-config

One thing that needs to be removed is adding commands to someones profile when you do make install

This should only be done with explicit permission (e.g. a user has enabled an option) if you like this option, but should definitely not be the default.

I'll update with more after I have reviewed the rest.

natis1:
Okay, first and most important of all I am going to make adding the commands to .profile an option and CERTAINLY not the default. I was mixed on even putting this in in the first place. As for an example cmake file I am not sure if it is necessary. However, in keeping with the standards set by settings.mk. I will move all user set settings to cmake-config and import in in CMakeLists. This will make editing it easier. Also I will change the directory that goma is made.

natis1:
I have completed my cmake-config file. I am working on two things now:
1) Append ${PROJECT_NAME}_ to all variables created by Goma. This will make the project more versitile and less prone to collisions. Also it will make it more like Trilinos.
2) Create a Readme.md (or else another markdown file) that accurately explains how to install Goma with cmake.

PR #118, opened 2016-07-06
Add cmake support, delete makefile, delete rf_conti.c, update readme …

natis1:
…to explain how to install Goma


CMakeLists.txt
```diff
@@ -0,0 +1,255 @@
...
+endif()
+
+
+if (${CMAKE_BUILD_TYPE} MATCHES DEBUG)
+    set(${${PROJECT_NAME}_EXTRA_C_FLAGS} "${${PROJECT_NAME}_EXTRA_C_FLAGS} -Wall")
+    set(${${PROJECT_NAME}_EXTRA_CXX_FLAGS} "${${PROJECT_NAME}_EXTRA_CXX_FLAGS} -Wall")
+    set(${${PROJECT_NAME}_EXTRA_F_FLAGS} "${${PROJECT_NAME}_EXTRA_F_FLAGS}")
+
+else()
+    set(${${PROJECT_NAME}_EXTRA_C_FLAGS} "${${PROJECT_NAME}_EXTRA_C_FLAGS} -c -Wall")
```
wortiz: You don't want -c here cmake should take care of that

-c creates object files in gcc

CMakeLists.txt
```diff
@@ -0,0 +1,255 @@
...
+#Run "ccmake ." to change any options here after running "cmake ."
+#option(${PROJECT_NAME}_Debug_Mode "This enables the debug flag in make, building it with Debug enabled." OFF)
+option(${PROJECT_NAME}_Has_Stratimikos "If you have stratimikos and want support for it enable this. This modifies the definitions" OFF)
+option(${PROJECT_NAME}_CXX11_STD "Unless you have an older compiler, or you used the NoCXX11 library, keep this enabled" ON)
+option(${PROJECT_NAME}_Write_To_Profile "This will automatically add the libraries to your profile, it is recommended to keep this off" OFF)
+
+
+#This adds a flag to the "User_Flags" section for stratimikos support. This flag is the same as the old makefile but I haven't tested it.
+if(${PROJECT_NAME}_Has_Stratimikos)
+    set(${PROJECT_NAME}_EXTRA_DEFINE "${${PROJECT_NAME}_EXTRA_DEFINE} -DHAVE_STRATIMIKOS")
```
wortiz: You might be able to set this based on whether we compiled trilinos with stratimikos or not

The trilinos cmake config has a trilinos package list that includes Stratimikos that you could look at

CMakeLists.txt
```diff
@@ -0,0 +1,255 @@
...
+#  "include/*.h"
+#  "include/brkfix/*.h"
+#  "include/brkfix/*.h"
+
+file(GLOB SOURCE_FILES
+  "src/*.cpp"
+  "src/*.c"
+  "src/*.F"
+  "src/*.C"
+  "src/brkfix/*.c")
```
wortiz: This needs the header files as well so they are dependencies in the makefile (i.e. we change something in a header file we want the source to be recompiled).

README.md
```diff
@@ -13,3 +13,46 @@ Changes in version 6.1
...
+
+###Get the Goma source code
+`git clone https://github.com/goma/goma && cd goma`
+This is by far the easiest way to get the source for Goma. It will download the latest source code of Goma.
+
+###Getting the prerequisites
+
+Goma requires many third party libraries which can be built automatically with the [build script](http://goma.github.io/build-script.html) located in the scripts directory.
+These are as follows:
+*[ARPACK](http://www.caam.rice.edu/software/ARPACK/): Arnoldi Solver package. A Fortran 77 program used by Goma to compute eigenvalues.
```
wortiz: Your markdown list items are rendering broken on github, most likely you need a space after the asterisk

CMakeLists.txt
natis1: I understand your concern but because it is listed as an include directory adding it to "SOURCE_FILES" is unnecessary and will not futureproof it since cmake will compile any necessary headers inside the directory automatically. Still, since it does no harm I have placed include/*.h and include/brkfix/*.h back in "SOURCE_FILES".

CMakeLists.txt
natis1: Maybe, I am looking into this right now. Still, someone may want to not compile stratimikos support even though they have stratimikos so they can make a goma executable others can use?? I don't know, but I will look into if we can remove this option

README.md
natis1: Yeah, I am seeing that, I think I fixed it.
* Like this instead

*of like this.

CMakeLists.txt
natis1: It will now default to whatever Trilinos_Stratimikos_FOUND is (this is a Trilinos variable). However it can be changed from this default by editing cmake-config or the cmake cache. Yes it compiles and works just fine like this.

wortiz:
Looks like you fixed my complaints and looks pretty good.

One issue that I forgot to mention is that we need to be able to build with seacas libraries that are separate from trilinos, as sometimes seacas github has bugfixes that haven't made it into trilinos releases

If you look at the settings.mk-example-seacas we have

    EXODUSII_LIB = -lexoIIv2c

    CHACO_LIB = -lchaco

    # NetCDF library (SEACAS version)

    NETCDF_LIB = -L$(SEACAS_TOP)/lib  -lnetcdf  -lhdf5_hl  -lhdf5

    # SEACAS includes and libraries
    SEACAS_INC = -isystem $(SEACAS_TOP)/include

    SEACAS_LIB = -L$(SEACAS_TOP)/lib  $(EXODUSII_LIB)  $(NETCDF_LIB) \
                   $(CHACO_LIB)

These are the libs that we need from seacas, mainly chaco and exodus (which relies on netcdf and hdf5)

This should be another option for cmake and we should be able to change what the exodusii library is called as sometimes it is libexoIIv2c.a and sometimes libexodus.a (or maybe this could be found by cmake)

Other than that I think we should have everyone try building from cmake with your updated readme and see what issues they run into.

cmake-config
```diff
@@ -33,23 +33,25 @@ if which cmake ; then
...
-    Trilinos_DIR=
-    MPI_DIR=
-    Sparse_DIR=
-    Arpack_DIR=
+    Trilinos_DIR=${Library_DIR}/trilinos-12.6.3-Built
+    MPI_DIR=${Library_DIR}/openmpi-1.6.4
+    Sparse_DIR=${Library_DIR}/sparse
+    Arpack_DIR=${Library_DIR}/ARPACK
+    ### Leave blank for autodetect. None needed if build seacas is off or not set.
+    SEACAS_DIR=${Library_DIR}/seacas
```
wortiz: You should probably revert these to blank as well as Library_DIR so your autodetect is default.

CMakeLists.txt
```diff
@@ -0,0 +1,277 @@
...
+### make uninstall support
+configure_file(
+    "${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_uninstall.cmake.in"
+    "${CMAKE_CURRENT_BINARY_DIR}/cmake/cmake_uninstall.cmake"
+    IMMEDIATE @ONLY)
+
+add_custom_target(uninstall COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_BINARY_DIR}/cmake/cmake_uninstall.cmake)
+
+### make debug will reconfigure with debugging enabled and then build in a debug directory
+add_custom_target(debug COMMAND "/bin/bash" ${CMAKE_SOURCE_DIR}/cmake-debug)
```
wortiz: I'm not sure I like this way of building debug executables

Maybe if cmake-debug 

was just something which called cmake-config -DCMAKE_BUILD_TYPE=Debug

That way your workflow would be

mkdir debug
cd debug
../cmake-debug

and then make / make clean as usual

natis1: Fixed and squashed into 1 commit.

natis1:
I consider this one basically perfect. It include an optional (but enabled by default) wrapper for goma which will export the relevant directories. You do indeed need to export them (I tested it), but it does work. The wrapper is installed only on make install and by default goes to goma/bin.

The wrapper WILL work anywhere but may require sudo if goma is installed to a location like /usr/local/bin which is owned by root.

cmake/runscript
```diff
@@ -0,0 +1,23 @@
...
+### Tells the file to add some temporary environmental variables
+echo 'export LD_LIBRARY_PATH='\"$2'/lib:$LD_LIBRARY_PATH'\" >> rungoma
+echo 'export PATH='\"$2'/bin:$PATH'\" >> rungoma
+echo 'export PATH='\"$3'/bin:$PATH'\" >> rungoma
+### Tells the file to run goma from anywhere
+### Just an ounce of cure against stupidity.
+echo 'cd '$1'/bin' >> rungoma
+### Tells this script to pass on any commands it gets to goma.
+echo './goma $@' >> rungoma
+
```
wortiz: Overall runscript comments:

We don't want to cd to the goma directory when running goma

you don't need to export PATH and LD_LIBRARY_PATH

you could have something like:

LD_LIBRARY_PATH=$MPI_DIR/lib:$LD_LIBRARY_PATH PATH=$SEACAS_DIR:$PATH $GOMADIR/goma

As one command and that would execute properly without the export.

One other thing which might be needed is adding seacas lib dir to path for standalone seacas as standalone seacas uses dynamic libraries.

natis1:
As an added bonus I fixed the bug where the continuous integration system would sometimes fail (~15% of the time) if the host was using the machine while a test was being run (it couldn't complete the shutdown routine, it was a VirtualBox bug) and sped up the continuous integration system by almost 30 seconds by forcefully terminating the VM. (this is okay because it immediately restores to a snapshot).

cmake-debug
```diff
@@ -0,0 +1,187 @@
...
+
+
+    ### These options will automatically add variables to your build flag
+
+    ### All blank options will be on if support is found and off otherwise.
+    ### This allows you to manually override that setting.
+    Build_Stratimikos=
+    ### Build with the standalone SEACAS instead of the Trilinos SEACAS
+    ### If this is off, Trilinos needs to be build with SEACAS.
+    ### Will automatically detect if seacas is standalone if left blank.
```
wortiz: This comment is incorrect from my testing.

So if I build trilinos without seacas and build seacas in the $GOMALIBS/seacas directory this does not get autodetected.

Maybe something needs to be done to correct the check for standalone seacas

(Changing to Standalone_SEACAS=ON compiled correctly so just the auto detect is broken)

natis1: You are correct, neither this nor the Stratimikos ones were working according to the comment. About to roll out a patch which fixes both.

natis1:
I did some testing and cmake is fully functional from 2.8.4 onward. Pre 2.8.4 and make debug will not accept options (like -j) but will accept definitions (like VERBOSE=1)

PR #120, opened 2016-08-01
Fixes a basis function bug for cross phase bcs

dshari:
This commit fixes an issue I ran into when using CROSS_PHASE boundary conditions with shells and continuum elements.


acochrane:
.buildgoma

wortiz:
This passed on my test suite.

I'll leave it unmerged for now so we can see why it isn't working in Jenkins

natis1:
.runguts

natis1:
.runguts

PR #128, opened 2016-11-10
Hook up existing relative liquid permeability model into shell porous…

tjiptowi:
… open flow equation

Add EXTERNAL_FIELD option in Rel Liq Permeability field


acochrane:
.runguts

PR #136, opened 2017-02-03
Fix checkfinite macro for newer compilers, add missing checkfinite guard

wortiz:
For when goma is compiled without checkfinite defined


acochrane:
.runguts

acochrane:
.runguts

PR #143, opened 2017-04-03
Put cross permeability evaluation in a separate function

tjiptowi:


wortiz:
Passes test suite and looks correct, are we having issues with jenkins?

wortiz:
.runguts

tjiptowi:
.runguts

PR #148, opened 2017-05-16
Finite Differencing upgrade & extraneous print

rbsecor:
Bugfix for AC conditions when x_AC is near zero but not zero, i.e. before FD steps on the order of 10E-200 were allowed resulting in singular AC matrices.  I consolidated most finite difference steps with a FD_FACTOR define in std.h.  Also removed an extraneous print statement from a previous commit.

GUTS results were the same except for p_lub_pore which now converges faster - good thing?

### DIFF
--- a/include/std.h
+++ b/include/std.h
@@ -233,6 +233,10 @@
 #ifndef DBL_MAX
 #define DBL_MAX 1.0E300
 #endif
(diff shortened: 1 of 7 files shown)


rbsecor:
.runguts

PR #152, opened 2017-07-06
Added Lame MU CONTACT_LINE2 model.

ravanga:
Based on the Lame CONTACT_LINE model, this adds a second region of localized mesh stiffening. It makes mu (shear modulus) become very large near two critical points, and decay to a lower value radially from those points. This helps keep elements near the critical points from shearing excessively while elements far away won't dilate much.

The model is applied in the material file under the "Lame MU" card as "CONTACT_LINE2" and has 7 inputs:
Lame MU = CONTACT_LINE2 {float1} {float2} {float3} {float4} {float5} {float6} {float7}
{float1} node set number for point one
{float2} minimum value of the shear modulus
{float3} maximum value of the shear modulus for point one
{float4} decay length for point one
{float5} node set number for point two
{float6} maximum value of the shear modulus for point two
{float7} decay length for point two

### DIFF
--- a/src/ac_update_parameter.c
+++ b/src/ac_update_parameter.c
@@ -728,15 +728,23 @@ update_MT_parameter(double lambda, /* Parameter value */
       break;
 
     case TAGC_LAME_MU_CONTACT_LINE_G0:
-      *(elc_glob[mn]->u_mu+1) = lambda;
+      elc_glob[mn]->u_mu[1] = lambda;
       break;
 
     case TAGC_LAME_MU_CONTACT_LINE_G1:
-      *(elc_glob[mn]->u_mu+2) = lambda;
+      elc_glob[mn]->u_mu[2] = lambda;
       break;
 
     case TAGC_LAME_MU_CONTACT_LINE_R0:
-      *(elc_glob[mn]->u_mu+3) = lambda;
+      elc_glob[mn]->u_mu[3] = lambda;
+      break;
+
+    case TAGC_LAME_MU_CONTACT_LINE_G2:
+      elc_glob[mn]->u_mu[5] = lambda;
+      break;
+
+    case TAGC_LAME_MU_CONTACT_LINE_R1:
+      elc_glob[mn]->u_mu[6] = lambda;
       break;
 
     case TAGC_LAME_LAMBDA:
@@ -1531,15 +1539,23 @@ retrieve_MT_parameter(double *lambda, /* Parameter value */
       break;
 
     case TAGC_LAME_MU_CONTACT_LINE_G0:
-      *lambda = *(elc_glob[mn]->u_mu+1);
+      *lambda = elc_glob[mn]->u_mu[1];
       break;
 
     case TAGC_LAME_MU_CONTACT_LINE_G1:
-      *lambda = *(elc_glob[mn]->u_mu+2);
+      *lambda = elc_glob[mn]->u_mu[2];
       break;
 
     case TAGC_LAME_MU_CONTACT_LINE_R0:
-      *lambda = *(elc_glob[mn]->u_mu+3);
+      *lambda = elc_glob[mn]->u_mu[3];
+      break;
+
+    case TAGC_LAME_MU_CONTACT_LINE_G2:
+      *lambda = elc_glob[mn]->u_mu[5];
+      break;
+
+    case TAGC_LAME_MU_CONTACT_LINE_R1:
+      *lambda = elc_glob[mn]->u_mu[6];
       break;
 
     case TAGC_LAME_LAMBDA:
--- a/src/mm_fill_solid.c
+++ b/src/mm_fill_solid.c
@@ -5059,16 +5059,16 @@ load_elastic_properties(struct Elastic_Constitutive *elcp,
 {
   /*local variables */
   int err;
-  int nnn, ins, inode, dim, dofs;
+  int nnn1, nnn2, ins, inode1, inode2, dim, dofs;
   int a, b, q, v, i, j, w;
 
   ELASTIC_CONST_STRUCT *elc_ptr;
 
   memset(d_mu_dx, 0.0, sizeof(double)*DIM*MDE);
 
-  elc_ptr = elcp;  
+  elc_ptr = elcp;
 
-  double dist, iie, value;
+  double dist1, dist2, iie, value;
 
   dim = ei->ielem_dim;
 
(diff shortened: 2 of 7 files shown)


src/mm_fill_solid.c
```diff
@@ -5072,39 +5072,151 @@ load_elastic_properties(struct Elastic_Constitutive *elcp,
...
+		                                / elc_ptr->u_mu[6];
+
+		  *mu = elc_ptr->lame_mu = elc_ptr->u_mu[1] + 0.1 / (pow(dist2,3.0)
+								     + 0.1 / elc_ptr->u_mu[5]);
+		}
+	    }
+	}
+      else
+	{
+	  if ( Proc_NS_Count[ nnn1] != 1 || Proc_NS_Count[ nnn2] != 1 )
```
tjiptowi: Do you need these lines

ravanga: Yes, these if statements check to make sure a valid NS number was used in the material card and help the user identify which one may be illegal.

src/mm_fill_solid.c
```diff
@@ -5072,39 +5072,151 @@ load_elastic_properties(struct Elastic_Constitutive *elcp,
...
+	  if ( Num_Proc == 1 )
+	    {
+	      EH(-1,"Cannot find second NS number on Lame MU CONTACT_LINE model");
+	    }
+	  else
+	    {
+	      if ( Proc_NS_Count[ nnn1] != 1 )
+		{
+		  if ( Num_Proc == 1 )
+		    EH(-1,"Illegal first NS number on Lame MU CONTACT_LINE model");
```
wortiz: These error exits that you have should happen in parallel as well shouldn't they (to be consistent)?

Otherwise it will just keep chugging along without any warning when you have NS with multiple nodes

src/mm_fill_solid.c
```diff
@@ -5072,39 +5072,151 @@ load_elastic_properties(struct Elastic_Constitutive *elcp,
...
+	
+	if (Proc_NS_Ids[ins] ==  (int) elc_ptr->u_mu[0]) nnn1 = ins;
+	if (Proc_NS_Ids[ins] ==  (int) elc_ptr->u_mu[4]) nnn2 = ins;
+      }
+
+      if ( nnn1 == -1 && nnn2 == -1 )
+	{
+	  if ( Num_Proc == 1 )
+	    {
+	      EH(-1,"Cannot find either NS numbers on Lame MU CONTACT_LINE model");
```
wortiz: Since your exiting when you can't find both nodesets here across the domain to be consistent in parallel you'd probably want to add a check somewhere before matrix fill fill to make sure those nodesets exist on some processor. Otherwise it looks like it would run as if they exist just not here in parallel

src/mm_fill_solid.c
ravanga: That's a good point, though this was designed with single node nodesets in mind.

src/mm_fill_solid.c
ravanga: You are right. In this case mu gets defaulted to 1 everywhere. My thinking is that any problem generally gets run in serial first as a check which would also check against this.

src/mm_fill_solid.c
```diff
@@ -5072,39 +5072,161 @@ load_elastic_properties(struct Elastic_Constitutive *elcp,
...
+
+      if ( nnn1 == -1 && nnn2 == -1 )
+	{
+	  if ( Num_Proc == 1 )
+	    {
+	      EH(-1,"Cannot find either NS numbers on Lame MU CONTACT_LINE2 model");
+	    }
+	  else
+	    {
+	      *mu = elc_ptr->lame_mu = elc_ptr->u_mu[1];
```
ravanga: Kris, I've corrected/added the code you were concerned about.

src/rf_setup_problem.c
```diff
@@ -1383,3 +1390,49 @@ find_MaxMatrlPerNode(void)
...
+ * Lame MU = CONTACT_LINE/2 models are present in the problem.
+ */
+
+{
+  int mn;
+  int length;
+  int ins;
+  int facticity = 0;
+
+  for (mn = 0; mn < (pd->Num_Species + 1); mn++)
```
wortiz: upd->Num_Mat; is probably what you want here

ravanga: You're right.

src/mm_fill_solid.c
```diff
@@ -5072,39 +5072,161 @@ load_elastic_properties(struct Elastic_Constitutive *elcp,
...
+		}
+	      else
+		{
+		  inode2 = Proc_NS_List[Proc_NS_Pointers[ nnn2]];
+
+		  dist2 = pow( (  pow(fv->x0[0] - Coor[0][inode2],2.0)
+			        + pow(fv->x0[1] - Coor[1][inode2],2.0) ), 0.5)
+		                                / elc_ptr->u_mu[6];
+
+		  *mu = elc_ptr->lame_mu = elc_ptr->u_mu[1] + 0.1 / (pow(dist2,3.0)
```
acochrane: Avoid using pow, it is slow.
Try to stick to `*`, `/` and `sqrt` for such operations, especially in assembly functions.

src/ac_update_parameter.c
```diff
@@ -739,6 +739,14 @@ update_MT_parameter(double lambda, /* Parameter value */
       *(elc_glob[mn]->u_mu+3) = lambda;
       break;
 
+    case TAGC_LAME_MU_CONTACT_LINE_G2:
+      *(elc_glob[mn]->u_mu+5) = lambda;
```
acochrane: I see you're picking up the present convention here, but indexing this way is the h-word.
Can you index into the array here like in your assemble function, i.e. `elc_glob[mn]->u_mu[5]`?

Also, what's up with the fifth element, `u_mu+4`? Is it set/used elsewhere?

src/rf_setup_problem.c
```diff
@@ -1383,3 +1390,49 @@ find_MaxMatrlPerNode(void)
...
+
+  for (mn = 0; mn < upd->Num_Mat; mn++)
+    {
+      elc = elc_glob[mn];
+      if (elc->lame_mu_model == CONTACT_LINE)
+	{
+	  length = dpi->num_node_sets_global;
+	  for (ins = 0; ins < length; ins++)
+	    {
+	      if (dpi->ns_id_global[ins] == elc->u_mu[0]) facticity = 1;
```
acochrane: Please avoid using single line `if` statements, they are not easier to read than multi-line `if () { ... }` blocks.

wortiz:
It looks like you need to perform the check for number of nodes inside your new function which uses dpi

dpi->ns_num_nodes_global should be what you need

There is an unlikely case where if you have some nodeset with say 2 nodes and those get broken into different processors your code won't catch that and will continue executing.

Highly unlikely that this would ever come up though.

wortiz:
Closing until issues are fixed.

src/mm_fill_solid.c
ravanga: Noted and changed throughout the function.

src/ac_update_parameter.c
ravanga: Noted and changed.

src/rf_setup_problem.c
ravanga: Understood and changed.

wortiz:
.runguts

acochrane:
.runguts

PR #153, opened 2017-07-13
Move setting of STRESS_CONT to index_post to after it is checked

wortiz:
This fixes a bug in the post processing "Stress contours". There was a check of STRESS_CONT == 2 after it was set to index_post.

This moves setting of STRESS_CONT to the initial index_post until after it is checked.

This came up in a multimode problem where index_post happened to be set to 2 when it got to STRESS_CONT, the code was executing as if STRESS_CONT was an external field in

                           if (STRESS_CONT == 2)
                             {
                               Export_XP_ID[index_post_export] = index_post;
                               index_post_export++;
                             }


wortiz:
.runguts

PR #155, opened 2017-07-31
Thin-film multiphase flow equations

acochrane:
Equations:
R_TFMP_MASS
R_TFMP_BOUND
Variables:
TFMP_PRES
TFMP_SAT

liquid outflow in SHELL_TFMP_FREE_LIQ_BC
Mass lumping
spurious oscillation control
added numerical diffusion integrated boundary condition
add inverse Peclet number postprocessing
added force from pressure integration post processing

### DIFF
--- a/src/mm_fill.c
+++ b/src/mm_fill.c
@@ -2068,7 +2068,10 @@ matrix_fill(
 	  err = assemble_shell_tfmp( time_value, theta, delta_t, xi, &pg_data, exo );
 	  EH( err, "assemble_shell_tfmp");
 #ifdef CHECK_FINITE
-	  CHECKFINITE("assemble_shell_tfmp");
+	  err = CHECKFINITE("assemble_shell_tfmp");
+	  if (err) {
+	    return -1;
+	  }
 #endif
 	}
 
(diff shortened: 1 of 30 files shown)


src/mm_fill.c
```diff
@@ -2062,6 +2063,15 @@ matrix_fill(
 
       	}
 
+      if(pde[R_TFMP_MASS] && pde[R_TFMP_BOUND])
+	{
+	  err = assemble_shell_tfmp( time_value, theta, delta_t, xi, &pg_data, exo );
+	  EH( err, "assemble_shell_tfmp");
+#ifdef CHECK_FINITE
+	  CHECKFINITE("assemble_shell_tfmp");
+#endif
```
wortiz: You should check the result of checkfinite here and return if there is an error (otherwise checkfinite can print gigabytes worth of text).

wortiz:
Some comments:

* There are a lot of new input cards added, adding some documentation might be good, even though we are in documentation limbo. I think traditionally CHANGELOG.tex was being used. I was also putting new cards into the docs folder

* There are a lot of hard coded material properties with hard coded units which seems to break from the rest of goma (I think I saw them in your assemble function and a few other places)

I didn't review the code that closely so I might've missed some other things

acochrane:
.runguts

acochrane:
.runguts

natis1:
.runguts

PR #158, opened 2017-08-22
Update Goma TPL to use newer versions of many libraries and much more.

natis1:
Complete changelog:

Create new build script to allow easy install of Goma with Intel compiler/mpi/mkl, and atlas math libraries

Streamline cmake script to make building debug versions of Goma (almost) as simple as with the make file. Remove cmake-debug.

Allow user to build Goma with their own copy of openMPI, such as the one typically located in /usr/lib

Allow user to build Goma TPLs with their own version of Cmake. This is highly recommended since cmake is recommended for building Goma as well.

Remove write to profile functionality from CMake. It should never have been included in the first place

Update Cmake

Update readme for new build script.

Create user-interaction script to query user on compiler, math libraries, and MPI

Fix typo in README.md | scripts/READE.md -> scripts/README.md

Add support for the following combinations of math libraries, compilers, and MPI:

You can now build Goma and its TPLs with the Intel compiler + any combination of MPI and Math Library. Preliminary functionality for standalone intel MKL (compiles successfully but has runtime errors with Trilinos). Standalone Atlas libraries for supercomputers and high performance PCs (although, due to Linux distribution filename differences you will have to specify these in a file unless your system is using atlas from the Ubuntu repos). Note: Using non-standard combination of libraries may result in unexpected results.

Other:

Update continuous integration VMs to use new version of openMPI.


natis1:
.runguts

wortiz:
Some comments:

In your README the script name is your old name for it (with capitalization)

I don't like the warning when cmake is not present, we added cmake to the script for redhat/centos 6 users who had older repo versions of cmake that trilinos did not support (I think they might be moving on to version 3 as well, which is past redhat/centos 7)

I also don't like the default build directory being inside the scripts directory, but I can't think of a better place to put it so that's probably fine.

Otherwise these look like some good improvements, especially being able to use existing MPI

natis1:
Ok. I fixed README.md and removed the warning. I also have allowed the use of built cmake for compiling Goma, although I have not tested this functionality.

scripts/build-goma-dep-trilinos-12.sh
```diff
@@ -16,187 +18,552 @@
...
----
-> #define AccessRoot $GOMA_LIB/SEACAS-2013-12-03
-47a48,50
-> #define ExcludeAnalysis
-> #define Parallel 0
-> 
-48a52
-> #define HasMatlab No
-"
+if cmake --version; then
```
acochrane: There are version of cmake that are too old for this build.
Cmake version should be checked and cmake should be built if version is too old.

Also, the version should only be checked once, setting some variable that is used later (see line 873)

wortiz: I don't know how bash short circuiting works but you could use something like this:

 [ $(cmake --version |grep "version" | awk '{print $NF}') \< "2.8.12.2" ]

When you know cmake exists

natis1: The answer lies in using sort to detect version number which, with -V, sorts versions properly, IE placing 11 after 3 instead of before it as in a traditional ascii sort. There's some code in easy-goma-builder that does something similar. I will add it to cmake as soon as I do some testing.

natis1:
Add support for cmake version detection and (hopefully this time) fix goma test bot so it actually displays resutls.

scripts/easy-goma-builder.sh
```diff
@@ -0,0 +1,232 @@
...
+
+CC_NAME="gnu"
+CC_CMD="gcc"
+COMPILER_VERSION=$(gcc --version | grep -o -e "[0-9]\+\.[0-9]*\.[0-9]* " | grep -o -e "[0-9]\+\.[0-9]*\.[0-9]*") &> /dev/null
+[ "$COMPILER_VERSION" = "`echo -e "4.8.1\n$COMPILER_VERSION" | sort -V | tail -n1`" ] && CXX11="true" || CXX11="false"
+if [[ "$CXX11" == "true" ]]; then
+   CXX11="true"
+else
+   CXX11="false"
+   exit 1
```
wortiz: Exits without message, maybe move your later message for this check to here

scripts/easy-goma-builder.sh
```diff
@@ -0,0 +1,232 @@
...
+fi
+MATH_LIBRARIES="netlib blas"
+SYS_LIB="-lm -lz"
+User_Flags="-O1"
+
+
+cd $(dirname `realpath $0`)
```
wortiz: realpath is a newer tool not installed/available on centos 6 coreutils, should probably use the && `pwd` method wherever realpath is involved

scripts/user-interaction.sh
```diff
@@ -0,0 +1,267 @@
...
+                    # Link intel libraries with gcc manually. See
+                    # https://software.intel.com/en-us/articles/intel-mkl-103-getting-started
+                    User_Flags="-O1 -m64 -DMKL_ILP64 -I${MKLROOT}/include"
+                    SYS_LIB="-lm -ldl"
+                fi
+            else
+                echo "But they will be ignored."
+                echo
+                echo "Blas will be downloaded and built from Netlib"
+                MATH_LIBRARIES="netlib blas"
```
acochrane: I picked netlib blas, but the goma executable is linked to the mkl in my parallel studio, resulting in an unrunnable goma executable.

PR #160, opened 2017-09-19
Sept 2017 catchup

rbsecor:
Catching up with some miscellaneous changes since the last Gomathon:

1. Transient supersedes Continuation in main.c
2. Updates to ROLL_FLUID_BC & VELO_SLIP_ROT_FLUID_BC for a thin coating layer.  These 2 BCs are still under construction.  Clean-up will be done when finalized.
3.  Beginnings of IDEAL_GAS expansion model for mesh for use with bubble problems

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -793,6 +793,12 @@ main(int argc, char **argv)
   /*
    *                           SOLVE THE PROBLEM
    */
+  if( TimeIntegration == TRANSIENT)
+    {
+        Continuation = ALC_NONE;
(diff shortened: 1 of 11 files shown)


src/main.c
```diff
@@ -793,6 +793,12 @@ main(int argc, char **argv)
   /*
    *                           SOLVE THE PROBLEM
    */
+  if( TimeIntegration == TRANSIENT)
```
wortiz: This statement should be moved after  the #ifdef DEBUG below to be consistent with the current debug run

PR #162, opened 2017-10-04
Get version from git and stamp exoII files with version and commit hash

acochrane:
Print version in intro text and in qa_records of exoII file


include/std.h
```diff
@@ -532,8 +538,15 @@ extern int zero_detJ_global;
  * versus 6 issue.)  This is not needed if goma is compiled with autoconf.
  * (Eric Benner 8/6/09)
  */ 
+
 #ifndef VERSION
-#define VERSION "6.0.0"
+
+  #ifdef GitVersion
```
wortiz: I would rather this be GIT_VERSION since its preprocessor

wortiz:
Other than my comment this looks fine as far as I can see.

Issue #149, opened 2017-05-18
Warnings under compile using gcc 5.4.0-ubuntu1~16.04.4 as of May 18th, 2017


acochrane:
The following is an inventory of files that generate compile warnings received as of May 18th, 2017 provided by the gcc-5.4.0 compiler.

These may not exist under other compilers, but I thought I would point them out for any interested parties to do some cleanup on.

bc_colloc.c               count : 1
rd_pixel_image2.c    count : 6
wr_exo.c                   count : 1
mm_augc_util.c        count : 1
mm_eh.c                  count : 2
mm_input.c               count : 3
mm_input_mp.c        count : 4
mm_post_proc.c       count : 1
mm_std_models.c    count : 1
ac_hunt.c                  count : 1
ac_particles.c           count : 1
brkfix/ppi.c                count : 5
brkfix/utils.c               count : 1

acochrane:
Addressed by #164 these warnings are no longer brought up at compile time.

PR #164, opened 2017-11-20
Fix SEACAS deprecated and some compiler warnings

wortiz:
Make compiling new things less annoying, we should add warnings to the jenkins bot so warnings don't build up again

There were also quite a few SEACAS functions that are now deprecated that I updated

**Requires GUTS branch fix-deprecated-and-warnings** as [MP] Weight Function's were being read in improperly and a required field was allowed to be left blank when SUPG was specified

### DIFF
--- a/src/ac_particles.c
+++ b/src/ac_particles.c
@@ -5639,12 +5639,16 @@ load_restart_file(void)
   zero_a_particle(&p);
   while(!feof(fp))
     {
+      char *fgetserr;
       for(i = 0; i < pdim; i++)
 	if(fscanf(fp, "%lf ", &p.x[i]) != 1)
 	  continue;		/* probable EOF */
       if(fscanf(fp, "%lf %d", &p.time, &p.owning_elem_id) != 2)
 	continue;		/* probable EOF */
-      fgets(garbage, 254, fp);
+      fgetserr = fgets(garbage, 254, fp);
+      if (fgetserr == NULL) {
+	EH(-1, "Error reading line in particle restart file");
+      }
       if(get_element_xi_newton(p.owning_elem_id, p.x, p.xi) == -1)
 	EH(-1, "Could not place particle from restart.");
       load_field_variables_at_xi(p.owning_elem_id, p.xi);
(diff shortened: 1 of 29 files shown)


wortiz:
Theres still some deprecated warnings that need to be fixed so wait on merging this until I get to them.

src/ac_particles.c
```diff
@@ -5639,12 +5639,16 @@ load_restart_file(void)
...
+      char *fgetserr;
       for(i = 0; i < pdim; i++)
 	if(fscanf(fp, "%lf ", &p.x[i]) != 1)
 	  continue;		/* probable EOF */
       if(fscanf(fp, "%lf %d", &p.time, &p.owning_elem_id) != 2)
 	continue;		/* probable EOF */
-      fgets(garbage, 254, fp);
+      fgetserr = fgets(garbage, 254, fp);
+      if (fgetserr == NULL) {
+	EH(-1, "Error reading line in partical restart file");
```
acochrane: Is this supposed to be particle?

src/mm_std_models.c
```diff
@@ -5924,7 +5924,13 @@ electrolyte_temperature (double t,         /* present value of time */
           t0 = tran->init_time;
         }
 #ifndef tflop
-            if(t <= t0) system("rm T_vs_t.out");  /* zero the file before writing to it */ 
+      if(t <= t0)
+	{
+	  int err;
+	  err = system("rm T_vs_t.out");  /* zero the file before writing to it */
```
acochrane: This seems uncomfortable. Does goma use system for removing other files?

src/ac_particles.c
wortiz: Ha, yes.

src/mm_std_models.c
wortiz: There are a couple places that goma uses system when it probably shouldn't

This one can probably be replaced with fopen write vs append

wortiz:
All the deprecated functions should be fixed with the last commit on this, should be ready to merge once it's been looked at

acochrane:
.runguts

acochrane:
This wraps up issue #149 
Thanks @wortiz !

PR #171, opened 2018-01-19
Update build-deps script to support CentOS 6

acochrane:
Add optional 'user' defined compiler paths in text of script


wortiz:
Appears to build properly and pass test suite with the build libraries.

Maybe add CXXSTD=-std=c++11 to the new settings.mk example

PR #180, opened 2018-06-11
Added EVSS_GRADV and LOG_CONF_GRADV as Polymer Stress Formulations

richmartin33:
Added two more options for Polymer Stress Formulation: EVSS_GRADV and LOG_CONF_GRADV. These Polymer Stress Formulations are versions of EVSS_F and LOG_CONF that use the velocity gradient in the assemble_stress functions instead of the velocity gradient projection G.

### DIFF
--- a/src/mm_fill_stress.c
+++ b/src/mm_fill_stress.c
@@ -1990,37 +1990,43 @@ assemble_stress_fortin(dbl tt,	/* parameter to vary time integration from
 						  if(pd->CoordinateSystem != CYLINDRICAL)
 						    {
 						      if( ucwt != 0)
-                                                       {
-                                                         for( k=0; k<VIM; k++)
-                                                           {
-                                                             advection_c -= ucwt*((double)delta(a,p)*bf[VELOCITY1+a]->grad_phi[j][k]*s[k][b] +
-                                                                                 (double)delta(b,p)*s[a][k]*bf[VELOCITY1+b]->grad_phi[j][k]);
-                                                           }
-                                                       }
-                                                     if( lcwt != 0.)
-                                                       {
-                                                         advection_c += lcwt*(s[a][p]*bf[VELOCITY1+b]->grad_phi[j][b] +
-                                                                              s[b][p]*bf[VELOCITY1+a]->grad_phi[j][a]);
-                                                       }
-                                                   }
-                                                 else
-                                                   {
-                                                     if( ucwt != 0)
-                                                       {
-                                                         for( k=0; k<VIM; k++)
-                                                           {
-                                                             advection_c -= ucwt*((double)delta(a,p)*bf[var]->grad_phi[j][k]*s[k][b] +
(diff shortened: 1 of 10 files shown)


src/mm_fill_stress.c
```diff
@@ -1683,14 +1691,30 @@ assemble_stress_fortin(dbl tt,	/* parameter to vary time integration from
 
       if( ucwt != 0. )
 	{
-	  (void) tensor_dot(s, g, s_dot_g, VIM);
-	  (void) tensor_dot(gt, s, gt_dot_s, VIM);
+	  if(evss_gradv)
```
wortiz: Rather than doing this for the residual portions I think it would probably be better just to load up g/gt with fv->G or fv->grad_v depending on the version

wortiz:
* Missing if check for Jacobian wrt G
* Missing correct Jacobian terms wrt velocity

src/mm_fill_terms.c
```diff
@@ -28117,7 +28117,7 @@ fluid_stress_conf( double Pi[DIM][DIM],
     }
 
   // Load up DEVSS-G flag and Gs
-  if(pd->v[POLYMER_STRESS11] && (vn->evssModel == EVSS_F || vn->evssModel==LOG_CONF))
+  if(pd->v[POLYMER_STRESS11] && (vn->evssModel==LOG_CONF || vn->evssModel==LOG_CONF_GRADV))
```
wortiz: I don't think you meant to remove EVSS_F here

wortiz: Nevermind I misread the function, I guess you could just remove evss_f in fluid_stress_conf as we always use the stabilization term for LOG_CONF formulations.

src/mm_fill_stress.c
```diff
@@ -1970,7 +1984,45 @@ assemble_stress_fortin(dbl tt,	/* parameter to vary time integration from
 						  advection_b *= R_advection;
 						}
 					      
-					      advection = advection_a +  advection_b;
+					      advection_c = 0.;
+					      if(evss_gradv)
+						{
+						  if(pd->CoordinateSystem != CYLINDRICAL)
```
wortiz: These if checks appear like they are unneeded, fluid stress loads all the derivatives at the same time so it does this as an optimization (not sure it actually saves much). Here you are just doing one Jacobian piece at a time.

Maybe I'm missing some other reasoning for them.

src/mm_fill_stress.c
```diff
@@ -1970,7 +1984,45 @@ assemble_stress_fortin(dbl tt,	/* parameter to vary time integration from
...
+						    {
+						      if( ucwt != 0)
+                                                       {
+                                                         for( k=0; k<VIM; k++)
+                                                           {
+                                                             advection_c -= ucwt*((double)delta(a,p)*bf[VELOCITY1+a]->grad_phi[j][k]*s[k][b] +
```
wortiz: These need to be updated to use grad_phi_e so that coordinate systems where grad(e_a) has values are accounted for.

PR #183, opened 2018-06-22
Compiler warning fixes

rbsecor:


wortiz:
.runguts

wortiz:
valgrind reports that there is usage of an uninitialized value in `fapply_moving_CA_sinh` on the velotheta problem

I can't really see where the uninitialized value is and valgrind usually isn't very helpful with stack variables.

This probably explains why it is passing on my machine but failing in the jenkins automated test

wortiz:
Looks like the memory issue has been fixed and matches the test suite when I reverted the rebaseline.

rbsecor:
I took a guess at what could be causing memory issues in what was added to the velo_theta BCs.  I noticed at least when I run valgrind it flags other issues particularly at the beginning of bc_integ.c...

wortiz:
Because of goma's egregious use of stack arrays/strtucts and depending on the size of MDE, MAX_PROB_VARS, MAX_CONC, etc.. valgrind usually ends up requiring us to use the max-stackframe option so that it doesn't think we are swapping stacks

Usually valgrind will print out a message saying to use --max-stackframe=<some size> but it can be hard to spot

Usually you can take that message and add a 0 to the end and you won't get anymore stackframe issues, probably not the best practice but until we switch to more heap allocated structs/arrays it works for now

For velotheta if I run with valgrind in the mess of valgrind issues I get:

    ==7120== Warning: client switching stacks?  SP change: 0xffea82400 --> 0xffefc24a0
    ==7120==          to suppress, use: --max-stackframe=5505184 or greater

So running valgrind with

    valgrind  --max-stackframe=55051840 ...

Will stop reporting all those errors that aren't really errors.

You can usually assume you need to use a larger stackframe when you see a ton of valgrind errors like that in goma as I've cleaned up most of the memory misuse issues (at least as far as the test suite is concerned).

PR #186, opened 2018-07-05
Add DPI information for rotated BCs

### LINKED ISSUE #132: Consistency Issues between side set associations in serial and parallel
`ss_to_blks` is used to associate a side set to a block id for applying boundary conditions. The association is not always the same in parallel causing BC's to be applied on multiple blocks

`SS_Internal_Boundary` is used to check if a sideset is internal. The logic for this is only for serial meshes.
In parallel it only takes into account the mesh for that processor when checking if it is an internal side set.
This causes SS_Internal_Boundary to be wrong depending on how the mesh was split by brk and then causes boundary conditions to be applied when they should not be. (e.g. KINEMATIC)

wortiz:
Adds new dpi variables to create ss_to_blks and SS_Internal_Boundary so that they are consistent with serial goma.

Allows the use of rotated BCs on double sided side sets in parallel.

Closes #132 

Linked to goma/brkfix#5

### DIFF
--- a/src/rd_dpi.c
+++ b/src/rd_dpi.c
@@ -149,14 +149,13 @@ static int get_variable_call_count = 0;
  * Prototypes of functions defined here and needed only here.
  */
 
-static void get_variable
-PROTO((const int netcdf_unit,
-       const nc_type netcdf_type,
-       const int num_dimensions,
-       const int dimension_val_1,
-       const int dimension_val_2,
-       const int variable_identifier,
-       void *variable_address));
+static void get_variable(const int netcdf_unit,
+                         const nc_type netcdf_type,
+                         const int num_dimensions,
+                         const int dimension_val_1,
+                         const int dimension_val_2,
+                         const int variable_identifier,
+                         void *variable_address);
 
 /**********************************************************************/
 /**********************************************************************/
@@ -667,9 +666,9 @@ rd_dpi(Dpi *d,
    *    for each type. NetCDF 2 wants to know all the dimensions!
    */
 
-  get_variable(u, NC_CHAR, 2, 
-	       d->num_elem_blocks_global,	d->len_string,
-	       si.eb_elem_type_global,	&(d->eb_elem_type_global[0][0]));
+  get_variable(u, NC_CHAR, 2,
+               d->num_elem_blocks_global, d->len_string,
+               si.eb_elem_type_global, &(d->eb_elem_type_global[0][0]));
 
   get_variable(u, NC_INT, 1, 
 	       d->num_elem_blocks_global,	-1, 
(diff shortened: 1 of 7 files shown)


wortiz:
.runguts

wortiz:
.runguts

wortiz:
.runguts

wortiz:
.runguts

wortiz:
.runguts

src/rd_dpi.c
```diff
@@ -1390,47 +1432,49 @@ get_variable(const int netcdf_unit,
...
+        }
+        break;
 
     default:
       EH(-1, "Specified netCDF data type unrecognized or unimplemented.");
       break;
     }
 #endif
 
+  if (err != NC_NOERR && is_required)
```
acochrane: Is it better to do the check for error inside the function or to hand in (return) a pointer where the error code/value gets stored?

src/rd_dpi.c
```diff
@@ -1443,37 +1487,60 @@ get_variable(const int netcdf_unit,
...
+static void
+get_variable(const int netcdf_unit,
+	     const nc_type netcdf_type,
+	     const int num_dimensions,
+	     const int dimension_val_1,
+	     const int dimension_val_2,
+	     const int variable_identifier,
+	     void *variable_address)
+{
+  get_variable_required(netcdf_unit, netcdf_type, num_dimensions, dimension_val_1, dimension_val_2, variable_identifier, variable_address, FALSE);
```
acochrane: Maybe make the wrapper handle more than just the one type of error?
Or at least handle the one but make it expandable.

src/rd_mesh.c
```diff
@@ -205,9 +205,17 @@ read_mesh_exoII(Exo_DB *exo,
     }
 	
 
+  // SS_Internal_Boundary uses the dpi values
```
acochrane: It's hard to cess out the true changes in a commit where style(spacing/tabbing) is updated in addition to code modification.

It is wise to separate style updates from code updates into different commits.

wortiz: Ignore whitespace: https://blog.github.com/2018-05-01-ignore-white-space-in-code-review/

src/rd_dpi.c
wortiz: It is probably better to return and propagate errors upwards, EH only sets an error flag and continues until we check for a parallel error.

I can probably add a return value for this fine so we exit rd_dpi/rd_mesh early

wortiz:
get_variable was only failing on one variable elem_var_tab_global, I made handling of this variable more appropriate (don't get if it doesn't exist) and now get_variable can error exit if any variables fail to be found.

This removes the wrapper function

PR #198, opened 2018-09-06
Fix makefile for isystem issue, fix some compiler warnings with newer gcc

wortiz:
Removes -I/usr/include from the trilinos_inc so that -isystem can be used. (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71090)

Fix some compiler warnings from the new defaults (mostly preventing overflow with sprintf).


wortiz:
@acochrane  Yeah it was from a misleading indentation someone might want to go through it and verify that's what should be done, but it looked like it should have been with the if statement.

I'm guessing someone forgot to add the brace or assumed the brace was there when adding a new term.

PR #208, opened 2018-10-31
Add a ENABLE_LOGGING ifdef, disable .log by default

### LINKED ISSUE #205: Disable .log files by default, put behind compiler define ENABLE_LOGFILES or something similar

wortiz:
Disables the .log files, to re-enable ENABLE_LOGGING can be passed as a compiler define.

Most people don't seem to look at these so it would be better to avoid creating them by default.

When running many problems in a single directory these can build up to large filesizes.

Closes #205

### DIFF
--- a/include/mm_eh.h
+++ b/include/mm_eh.h
@@ -118,10 +118,27 @@ extern int current_severity;	/* global error signal (-1=die,0=prnt,1=dbg) */
  *
  * Requires use of a local static const char yo[] ="Routine_name".
  */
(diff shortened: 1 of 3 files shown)


wortiz:
It looks like it is just hidden under the expand code button for the diff.

wortiz:
If you look here it is easier to see https://github.com/goma/goma/blob/778196cc96a86857e98eb6ef15c2b1d7a3d9e24e/src/mm_eh.c#L162

PR #211, opened 2018-11-19
Thanksgiving

rbsecor:
Cleaned up the logarithmic hunting capability - i.e. for parameters that vary by more than an order of magnitude, continue in the logarithm of the parameter instead of the parameter itself.  Now, logarithmic hunting is set using the "ramp" variable with hunt[IHC].ramp = 2.

So the LOG_HUNTING_PLEASE compiler define is not needed, designate linear or logarithmic via the input file.

Also, other clean-up in U_PARABOLA, etc. (some missing "break"s)


wortiz:
I re-added the ldc-hunt back to the test suite it seems like it was working fine, no one rebaselined it.

You should be able to pull master for guts and get the update. Otherwise it should run on this next check.

It appears that your update breaks the step size partway through the ldc-hunt test problem.

Perhaps from this removal?

https://github.com/goma/goma/pull/211/files?utf8=%E2%9C%93&diff=unified&w=1#diff-4d163264545d3f876b1e553f3b1a84d3L1434

Not sure though.

rbsecor:
I think I would be in favor of rebaselining the out.exoII.  The current version completes in 15 steps compared to 20 (steps 1-10 are identical - the old version takes smaller steps starting at step 11) - and I can't pick out any difference in the solutions.

acochrane:
Does this mean changing the default hunting step size control to logarithmic?

rbsecor:
No. The logarithmic stuff is not invoked in any test suite problem.

wortiz:
I looked at it a little bit more. It seems like the issue is that the test suite problem is using the constant step size (ramp == 1) but your update changes it to use the step control.

I think we probably want to keep the constant step size for some cases.

If you run through a debugger and watch delta_s[0] and delta_s[1] you can see the old code only sets the values once on update_parameterHC ac_hunt.c:473 

Whereas the update updates delta_s every step according to:

https://github.com/goma/goma/blob/65bf62e28dc164f9a61fe65b310acd6b5fd2baea/src/ac_hunt.c#L773

And somehow it is getting into (looks like because it isn't in an iHC=0;iHC<nHC;iHC++ loop

https://github.com/goma/goma/blob/65bf62e28dc164f9a61fe65b310acd6b5fd2baea/src/ac_hunt.c#L1277

to change dhunt_par_new with path_step_control.

src/mm_post_proc.c
```diff
@@ -807,19 +807,19 @@ calc_standard_fields(double **post_proc_vect, /* rhs vector now called
...
+      for (b = 0; b < dim; b++) {
+	Dsh[a][b] = (fv->grad_v[a][b] + fv->grad_v[b][a]);
       }
     }
     /* find second invariant of strain-rate */
     Dnn = 0.0;
-    for (a = 0; a < VIM; a++) {
-      for (b = 0; b < VIM; b++) {
-	Dnn += 0.5 * (Dsh[a][b] * Dsh[a][b] - Dsh[a][a] * Dsh[b][b]);
+    for (a = 0; a < ; a++) {
```
wortiz: Won't compile because of missing dim here and the next line.

I'm guessing these should be kept as the VIM you changed it too though. As when VIM is 3 for cylindrical, swirling, etc... you can get gradients involving the third coordinate.

PR #212, opened 2018-12-07
Fixed error in two include compiler directives.

DrBobF:
Removed the extra spaces within the #include directives


wortiz:
.runguts

PR #216, opened 2018-12-14
Shear

rbsecor:
Looking more at the MEAN_SHEAR post-processing calculations, the loop index needs to go to VIM.  Otherwise, for CYLINDRICAL problems the [2][2] component - V_r/r - will not be included.  This breaks the velotheta test problem exoII comparison for SHEAR values but VELOCITY components.  Ergo, this test problem will need to be re-baselined.


wortiz:
.runguts

PR #223, opened 2019-01-07
Clean up #2

DrBobF:
Added some more POSIX declare statements
Added some guards to prevent some multiple inclusion of a couple of header files
Changed the GOMA's GIT VERSION printing - didn't work correctly
Minor clean up


wortiz:
.runguts

wortiz:
.runguts

wortiz:
.runguts

include/user_shell.h
```diff
@@ -9,6 +9,9 @@
 *                                                                         *
 * This software is distributed under the GNU General Public License.      *
 \************************************************************************/
+
+#ifndef _USER_SHELL_H
+#define _USER_SHELL_H
```
wortiz: These preprocessor definitions for this file I think should be removed this like the defn files isn't really a header file it is just c source to be added in place of this:

https://github.com/goma/goma/blob/master/src/mm_fill_shell.c#L778

Ideally it shouldn't be done this way, I'm not seeing how this is any more convenient than just having the code in mm_fill_shell.c but that can be fixed at a later point.

DrBobF: I agree. Now that I'm looking, I don't even see user_shell.h in the Makefile. Perhaps someone that runs "shell surface charge" problems can give some insight. I think what probably happened was someone was testing different versions of this section of code and forgot to paste the final version into mm_fill_shell.c

wortiz:
.runguts

PR #224, opened 2019-01-08
Corrected finding the max_force on particles

DrBobF:
Corrected finding the max_force on particles during the adjustment of the time step.


wortiz:
.runguts

PR #235, opened 2019-02-20
More warning cleanup

wortiz:
Mostly missing break statements or adding explicit /* fall through */ comment so compiler doesn't complain about falling through to next case.


wortiz:
.runguts

PR #239, opened 2019-02-27
Suspension balance push

richmartin33:
This includes many changes to the suspension balance model. The main features include (1) a rotated Q-tensor, (2) retrieving the velocity gradient from the previous time step in the function divergence_particle_stress, (3) a few new post processing variables, (4) an addition of the non-local shear rate, and (5) an optional input parameter "Suspension Balance Length Scales" that takes the particle radius, characteristic length scale, and maximum velocity as inputs.


src/mm_viscosity.c
```diff
@@ -1974,7 +1974,7 @@ suspension_viscosity(int species, /* species for solid volume fraction track */
      this may be too low and I will have to adjust it.
      */
 
-  if ( nexp > 0.0 || ( C > 0.0 && C < 0.90 * maxpack)  )
+  if ( nexp > 0.0 || ( C > 0.0 && C < 0.98 * maxpack)  )
```
acochrane: Is there any more guidance on this, 10% vs 2% or is it fine?

include/mm_mp_const.h
```diff
@@ -71,6 +71,8 @@ extern int Num_Var_Init_Mat[MAX_NUMBER_MATLS];	/* number of variables to overwri
 #define EM_DISS         21   /* heat generation by EM waves */
 
 
+/* Switch for one of the the suspension balance model inputs */
+extern int SBM_LENGTHS;
```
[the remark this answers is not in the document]
wortiz: I'll probably move this.

include/mm_std_models.h
```diff
@@ -120,10 +120,15 @@ PROTO((dbl [DIM],               /* divergence of the stress*/
        dbl [DIM][MAX_CONC][MDE],/* derivative wrt concentration */
        dbl [DIM][DIM][MDE],     /* derivative wrt velocity */
        dbl [DIM][DIM][MDE],     /* derivative wrt mesh */
-       dbl [DIM][DIM][MDE],     /* derivative wrt vorticity direction */
+       dbl [DIM][DIM][DIM][MDE],   /* derivative wrt velocity gradient G */
```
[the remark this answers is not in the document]
wortiz: It looks like this was changed to use a different formulation and currently isn't used so it could probably be removed

src/mm_fill_species.c
```diff
@@ -368,24 +371,7 @@ assemble_mass_transport(double time, /* present time valuel; KSC             */
...
-	  h_elem_inv=0.;
-	}
-      else
-	{
-	  h_elem_inv=1./h_elem;
-	}
-	
-    }
+
   /* end Petrov-Galerkin addition */
```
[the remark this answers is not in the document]
wortiz: This can be removed as supg is now per species

src/mm_fill_terms.c
```diff
@@ -29497,7 +29499,7 @@ fluid_stress( double Pi[DIM][DIM],
         }
     }
 
-  if ( d_Pi != NULL && pd->v[VELOCITY_GRADIENT11] )
+  if ( d_Pi != NULL && pd->v[VELOCITY_GRADIENT11] && pd->v[POLYMER_STRESS11])
```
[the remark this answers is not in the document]
wortiz: G is only used for the DEVSS stabilization so this is probably more correct

src/mm_viscosity.c
wortiz: This is a cutoff to keep concentration between 0-1 so it's probably okay to keep for now.

wortiz:
Note before merge: Species SUPG needs mesh derivatives.

wortiz:
SUPG mesh derivatives added and comments addressed.

src/mm_fill_species.c
```diff
@@ -114,6 +114,121 @@
 /******************************************************************************/
 /******************************************************************************/
 
+struct SUPG_terms {
```
acochrane: Seems like this would be next to the definition of the PG_DATA struct.

src/mm_fill_species.c
```diff
@@ -114,6 +114,121 @@
...
+  dbl supg_tau;
+  dbl d_supg_tau_dv[DIM][MDE];
+  dbl d_supg_tau_dX[DIM][MDE];
+};
+
+void get_supg_tau(struct SUPG_terms *supg_terms,
```
acochrane: And this would be in a util file, but to be fair, the pg_data struct is filled with code in the matrix fill function.

src/mm_fill_species.c
wortiz: Moved.

src/mm_fill_species.c
wortiz: Moved to a utility file

PR #246, opened 2019-03-27
2.5D Cartesian

rbsecor:
Adding a 2-1/2D Cartesian capability invoked with CARTESIAN_2pt5D.  Z-direction pressure gradient is specified through the Navier Stokes Source card.

Also adding ALLOW_NEGATIVE_TIMES_PLEASE define (self-explanatory I think).


wortiz:
.runguts

PR #250, opened 2019-05-14
Build script typo

DrBobF:


wortiz:
That is checking for intel mpi c compiler which is mpiicc to match their icc c compiler.

DrBobF:
Well, something is not quite. Let's say we find the directory /usr/lib64/openmpi or /usr/lib/openmpi then MPI_BASE_DIR is set appropriately.  Then if we do not have the Intel mpiicc the MPI_BASE_DIR variable is reset to BUILD.  Is that the correct behavior?

wortiz:
I understand that now, then maybe the fix is to remove that line completely as it doesn't seem to be doing anything useful.

DrBobF:
The code  is checking to see if mpiicc is available. In this case, shouldn't MPI_BASE_DIR be set to the location of the Intel mpiicc? So mpiicc if-statement should be patterned like the openmpi version; checking the existence of the directory. I don't know where mpiicc is typically installed? Now where to place this mpiicc if-statement is tricky. Perhaps a case statement based on MPI_NAME? Then checking existence of directories with MPI_BASE_DIR="Build" as the default? Maybe the "recommended goma TPL settings" if-statements below need to encompass the directory checks? Which one do you think is cleaner and easier to maintain?

wortiz:
So from what I remember is that intel needs to be explicitly enabled by hitting no as we never got intel builds working well which I guess is why it is setup the current way (though that if check is literally only causing the reset issue)

So I think the correct action would probably be to remove:
    if which mpiicc &> /dev/null; then
        MPI_BASE_DIR="BUILD"
    fi

Then I think the correct behavior would be preserved (suggesting using openmpi, then building)

It looks like we set to `MPI_BASE_DIR="intelMPI"` for intel but I don't have easy access to intel compilers here to test different setups.

DrBobF:
In the easy-goma-builder.sh script, 
if we don't find an openmpi directory or, if we keep the if-statement in question, "which mpiicc" doesn't return true - we set MPI_BASE_DIR is set to BUILD 

But in the build-goma-dep-trilinos-12.sh script (337),
there is a comment that suggestions there is a conflict if the use requests intel mpi AND sets the MPI_BASE_DIR="BUILD".

So, it looks like IF a user wants the Intel MPI it has to be pre-built by the the user and the MPI_BASE_DIR has to be specified by the user, else the script doesn't provide the correct action (as written).

Maybe, like the TPLs it would be simpler logic and easier to maintain if we break apart the scripts; one for Intel based stuff and one for everything else? Especially since you mentioned that the Intel builds were always problematic.

wortiz:
First a question:
Are you trying to build with intel, or a built in openmpi, or just fixing issues?

Second there are a couple issues.

1. I think the logic for easy-goma-builder is a bit off, looking at it with @acochrane we can understand why it was setting it to build because otherwise it would end up using the intel mpi compilers that are first on the path but it would try to set the base dir to that of openmpi if it was found.

2. easy goma builder seems to be very finicky with what users already have on their machine I think @acochrane and I have found the best way to build libraries is just to edit lines 25 to 77 in https://github.com/goma/goma/blob/master/scripts/build-goma-dep-trilinos-12.sh#L25-L77 manually and forgo the "easy builder"

> Maybe, like the TPLs it would be simpler logic and easier to maintain if we break apart the scripts; one for Intel based stuff and one for everything else? Especially since you mentioned that the Intel builds were always problematic.

That could be a possibility 

I think we were talking about maybe just doing the compiler test, mpi test in place of those lines at the start which are causing issues and then using those results for the tpl settings

DrBobF:
I'm trying to build with openmpi.

It's difficult to manage all of the different variables that need to be changed in different places, in different scripts and CMakeLists to keep everything consistent (pointing to the same versions/locations) throughout the build process without keeping a logbook.  OpenMPI seems to be one of those problematic TPLs.

I was trying to really understand the logic between easy-goma-builder, build-goma-dep and the functions within the user-interaction.

I've resorted to using variables for **all** of the TPLs throughout like:

CMAKE_VERSION="3.9.5"
CMAKE_MD5="0922130d0e0c142a88e58c6e4fef4d7d"
OPENMPI_VERSION="2.1.3"
OPENMPI_MD5="2d4190db760ab15c6efd812079342101"

so I only have to change them in a single place. I was glad to see you start doing that in the recent release of the script.  This way I don't have to maintain a local version and merge the "official" version with my local one.

wortiz:
> I was trying to really understand the logic between easy-goma-builder, build-goma-dep and the functions within the user-interaction.

easy-goma-builder was made to be a wrapper around the build-goma-dep script that automatically set variables for you, though I'm not sure how much of a success it was. user-interaction is just a bunch of utils to make that happen.

DrBobF:
I understand the purpose and the interactions of the scripts; just need to make sure that I fully understand the logic within each -- I've been trying to cut out what I don't need to make it simpler to swap in and out different packages.

PR #261, opened 2019-07-23
Fillet update

rbsecor:
Updated FILLET_BC to include chamfers and a little cleanup
Connected post-processing MEAN_SHEAR to calc_shearrate routine


wortiz:
VELOTHETA needs to be rebaselined to update to the mean shear post processing, talking with @tjiptowi  this appears to match Bird et al

Any idea where the 4 comes from in the manual?

Screenshot from 2019-07-24 09-21-57

rbsecor:
There are 2 commonly used sets of tensor invariants.  The set in Bird et al. is based on the traces of the matrix, the square, and the cube.  The other set is based on the first, second, and third order trace of the matrix.  The 2 sets are related - see DPL, Volume 1, pp. 568-9.  Beyond this, the scalar shear rate is generally agreed to be the square root of shear rate tensor (second order trace of velocity gradient tensor plus transpose.  The D tensor above is 1/2 velocity gradient tensor plus 1/2 the transpose.  So, a factor of 2 (square root of 1/2 squared) from D versus shear rate tensor and another factor of 2 from Brid style invariants to the more commonly used ones.

In this check-in, I tied the calculation of MEAN_SHEAR to the calc_shearrate subroutine to avoid the possibility of different definitions in multiple places in Goma.

wortiz:
.runguts

PR #266, opened 2019-09-06
Coloring for stress finite difference

wortiz:
Improves log conf speed by about 40%, not sure why it wasn't added before


wortiz:
Added a few bugfixes that never got commited

PR #271, opened 2019-11-20
Mixing

rbsecor:
Added some post-processing variables that are reported to be useful in quantifying degree of mixing ; helicity, Lamb vector, etc.
Also made the hunting exodus time step always positive in order to make paraview work better.


wortiz:
I'm going to merge this as it is more robust than what is currently happening

Kris and I had discussed this issue with output and had decided that the best possible way to get around this would to make the exodus time steps in numeric order e.g. 1,2,3....N and save the continuation/hunting values in exodus global variables (like we currently do with things like CONVRATE and AC values), this could also make restarting a bit better (at least for hunting) as we would be able to get all values instead of just the first one (in the hunting case).

PR #272, opened 2019-12-06
fix bug in load dofptrs

wortiz:
Not sure if this caused any bugs but could cause failures or garbage reads if `peqn` was the 0th equation for these


wortiz:
Nevermind not actually a bug misread, was causing a bug on segregated branch due to how pd->e is created

PR #279, opened 2020-02-20
Feb2020

rbsecor:
Some viscoelastic updates.  Addition of a VE Stress condition to go along with U_PARABOLA to test against STRESS_DEVELOPED.  Some post-processing additions.

TCOND[12] was changed to TCOND[XY] for compatibility with Paraview and other post-processors.  This change will require a guts re-baseline.


src/mm_post_proc.c
```diff
@@ -565,7 +565,7 @@ calc_standard_fields(double **post_proc_vect, /* rhs vector now called
   struct Species_Conservation_Terms s_terms; 
 
   /* DG VE stuff */
-  int v_s[MAX_MODES][DIM][DIM];
+  int v_s[MAX_MODES][DIM][DIM], ves[DIM][DIM];
```
wortiz: ves is used as a double array so this should be changed

rbsecor: Done

PR #283, opened 2020-03-03
VE_update_march

rbsecor:
Some axisymmetric fixes for Log conformation method.  Also, the beginning part of an analytical eigenvalue Jacobian routine - not yet operational, currently undergoing testing.  Slight updates to the numerical Jacobian routine for the conformation tensor


src/mm_fill_stress.c
```diff
@@ -5802,74 +5806,250 @@ compute_exp_s(double s[DIM][DIM],
...
  // Solve for R and eigenvalues of conformation tensor (exp_s)
 
- // convert to column major
+  for (i = 0; i < VIM; i++) {
+    eig_values[i] = exp(W[i]);
+  }
   for (i = 0; i < VIM; i++) {
     for (j = 0; j < VIM; j++) {
-      A[i*VIM + j] = exp_s[j][i];
+      R[i][j] = U[i][j];
```
wortiz: This is a pretty big change, are you seeing similar behavior as to before?

Fattal and Kupferman used the conformation tensor (exp_s) to find rotations while this is using the log c (s)

Screenshot_2020-03-03_08-44-31

wortiz:
> Slight updates to the numerical Jacobian routine for the conformation tensor

I didn't see these did they get missed in the commit?

rbsecor:
Yes, same behavior.  This makes use of the spectral decomposition theorem which is well-known.

Robert

src/mm_fill_stress.c
wortiz: Thanks, it makes sense, I guess we are using it anyway to take the exponent of a matrix so the eigenvectors should be equivalent

src/mm_fill_stress.c
```diff
@@ -5815,6 +5815,9 @@ compute_exp_s(double s[DIM][DIM],
     }    
   }
 
+#if 1
```
wortiz: You probably should switch these off or put a #if DEBUG_LOGC or something for the merge

src/mm_fill_stress.c
```diff
@@ -5981,7 +5984,7 @@ analytical_exp_s(double s[DIM][DIM],
     }
   }
 
-#if 0
+#if 1
```
wortiz: Second one

rrrao:
An equivalent mathematic formulation may not have the same behavior, which Rich tried to show in his blade coating paper. Log c and c should be equivalent but aren’t. I guess we need some test suite coverage?

wortiz:
@rrrao We have test suite coverage of log conf, in this case there should be no mathematical difference in formulation. We were just doing extra work for no reason.

s = QDQ^-1
exp_s = Q exp(D) Q^-1

What @rbsecor fixed was not recalculating the decomposition to get Q for exp_s since we already had it

rrrao:
Good to know – thanks Robert. The log conf stuff is not as robust as we had hoped and only the original form has good We limits.

PR #287, opened 2020-03-27
Replace ftp-link to netcdf package to https-link

AndrewUshakov:
In corporate environment it is frequently impossible to download anything (in our case - netcdf package) from ftp. Working alternative https-link has been found on the page [NetCDF Downloads](https://www.unidata.ucar.edu/downloads/netcdf/) and is suggested now as a replacement.


wortiz:
.runguts

wortiz:
Thanks for the updated link.

PR #289, opened 2020-04-09
Sync hdf5 file name in ARCHIVE_NAMES with ARCHIVE_URLS

AndrewUshakov:
Replace invalid hdf5 file extension ".tar.gz" in  ARCHIVE_NAMES, used to verify downloads, by the expected ".tar.bz2" extension from ARCHIVE_URLS  in the script build-goma-dependencies.sh.

### DIFF
--- a/scripts/build-goma-dependencies.sh
+++ b/scripts/build-goma-dependencies.sh
@@ -240,7 +240,7 @@ MATIO_MD5="170075cce5c144e19f610af9b64cb63b"
 
 ARCHIVE_NAMES=("arpack96.tar.gz" \
 "patch.tar.gz" \
-"hdf5-${HDF5_VERSION}.tar.gz" \
+"hdf5-${HDF5_VERSION}.tar.bz2" \
 "netcdf-${NETCDF_VERSION}.tar.gz" \
 "parmetis-4.0.3.tar.gz" \
 "sparse.tar.gz" \


wortiz:
.runguts

wortiz:
This shouldn't have affected any checking, we were just downloading to a file with the wrong extension, tar is thankfully smarter than we are.

Thanks for the fix

AndrewUshakov:
I saw a message about wrong md5 sum of absent .tar.gz while valid tar.bz2 was previously copied,  pressed 'c' to continue and script tried to download openmpi from internet. Unsuccessfully, because of ntlm2 only authentication on proxy, absence of admin rights, etc.

wortiz:
I just checked again with the tar.gz and I don't get any issues, it just names the bz2 as gz

Sadly we don't have a lot of error checking in the script so probably a bad wget or some other issue caused it. I have run into a glitch where it tries to download to the home directory when certain paths are given like "~/gomalibs" rather than a full path that can cause some issues.

If you are running the script on a few machines you can always reuse the tars folder or just download links manually and place in a tars folder where you want to build the gomalibs, I know a few people had issues on their corporate clusters running the script due to restrictions of network access.

At least with your fix here everything should be the appropriate name in the download links.

AndrewUshakov:
Today morning I was able to reproduce the error.

1) I copied GOMA distributive with already downloaded files to the tars subdirectory to the isolated server (without openmpi, because I have it installed in my virtual machine).
2) Started build script,
3) Script determined that installed openmpi is outdated, tried to download and I stopped it.
4) I, manually on WIndows machine, downloaded openmpi using url, found in ARCHIVE_URLS  from the build script:

https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.10/hdf5-${HDF5_VERSION}/src/hdf5-${HDF5_VERSION}.tar.bz2

and copied it (with the original tar.bz2 extension - only now I realized, that script save files not with the original names, but with the names from ARCHIVE_NAMES) to the tars sub directory on the server. 

5) Cleaned up neighbor sub folders and restarted buid script. Result is:

`
Check for openmpi-4.0.2.tar.gz at https://download.open-mpi.org/release/open-mpi/v4.0/openmpi-4.0.2.tar.bz2
--2020-04-10 10:18:47--  https://download.open-mpi.org/release/open-mpi/v4.0/openmpi-4.0.2.tar.bz2
Resolving download.open-mpi.org (download.open-mpi.org)... 13.226.77.118, 13.226.77.123, 13.226.77.120, ...
Connecting to download.open-mpi.org (download.open-mpi.org)|13.226.77.118|:443... connected.
Unable to establish SSL connection.
Issue checksum with archive:
2fdfb4d31cfe1363d8bb5e7488e57cdf
d41d8cd98f00b204e9800998ecf8427e
openmpi-4.0.2.tar.gz
Enter "c" to continue (any other letter to exit):
ctar: This does not look like a tar archive

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
`
Then I properly corrected error on the server version of build script, but prepared pull request, which corrects similar error with hdf5 library ;) I'll prepare second pull request soon.

PR #290, opened 2020-04-10
Sync openmpi file name in ARCHIVE_NAMES with ARCHIVE_URLS

AndrewUshakov:
Correct openmpi file extension in ARCHIVE_NAMES to avoid issue with manually downloaded and already copied to the tars sub directory openmpi distributive with the original file name (extension). See additional explanations when exactly error occures in [https://github.com/goma/goma/pull/289](url).


wortiz:
.runguts

PR #291, opened 2020-04-17
Bond viscosity

rbsecor:
Tweaks to the BOND_EVOLUTION viscosity model as requested by Eric Vandre.  I assume we did not yet have a BOND problem in the test suite?


wortiz:
Yeah no bond evolution in test suite.

PR #297, opened 2020-05-29
Add Saramito model to Goma

jtmcconnell:
This pull request adds the Saramito yield model to some of Goma's existing viscoelastic models, specifically, the Oldroyd-B, Giesekus and PTT models. To use the Saramito model, one should use one of the following specifications for the _Polymer Constitutive Equation_ input card:

SARAMITO-OLDROYDB
SARAMITO-GIESEKUS
SARAMITO-PTT

Additionally, a  _Polymer Yield Stress_ card is necessary when using Saramito model.


wortiz:
.runguts

wortiz:
.runguts

wortiz:
The last one failed because convergence rate changed slightly on a problem most likely due to order of operations changes, I've reduced tolerance on convergence rate for that problem. Should now pass

    
    CONVRATE     rel diff:  2.1418082e+00 ~  2.1445412e+00 = 1.27437e-03 (FAILED)

PR #300, opened 2020-06-30
Stamp_rollon

rbsecor:
Bringing user routines of some micro-printing stamp geometry and velocity routines into main goma, albeit in the form of include files at this point only available from me.

Also some updates to the CAPILLARY_TABLE bc which allows variable pressure to be applied to a free surface such as an air jet or knife.  Fixes to setup_table_bc and other routines were required since the table scale factor is not always the first BC float, i.e., it isn't for CAPILLARY_TABLE


src/bc_colloc.c
```diff
@@ -1081,6 +1091,10 @@ fprintf(stderr,"circle %g %g %g %g\n",xcirc,ycirc,xcen2, ycen2);
 } /* END of routine f_double_rad                                             */
 /*****************************************************************************/
 
+#ifdef FEATURE_ROLLON_PLEASE
+#include "feature_rollon.h"
```
wortiz, line 1095: Is this a file you wanted to check in? If so you need to add it with git add

rbsecor: There's that file and one other that, for now, I need to keep off github.  i.e., is currently considered 3M confidential.  I can provide these files to those with a "need-to-know" for work under the Printability CRADA.

wortiz: Okay just wanted to verify you didn't accidentally forget it

PR #302, opened 2020-07-02
apply test function to yflux_user

wortiz:


wortiz:
I haven't heard back if this is necessary for anything, as it seems to be something left in when debugging I'm going to merge for now

PR #303, opened 2020-07-17
LOG CONF zero evalue

rbsecor:
Added bulk shear stress term to the CAPILLARY card in addition to bulk pressure.  Also experimented with zero eigenvalue checks in assemble_stress_log_conf.  I believe these changes produced marginal improvement in testcases.  I attempted to limit the bulk of error messages when the LOG_CONF source term goes infinite, but gave up after awhile.


wortiz:
Still looking it over but the failure was from the WH added to CAPBC in p_fibre

rbsecor:
OK, I was trying to find out which one … my normal workstation gives multiple spurious exoII comparisons diffs.

Robert

wortiz:
Alright I've taken a bit more of a look at it.

A couple notes 

Comparing floating point 0.0 with == is technically okay when you are comparing a manually set value, but I'm okay with changing them to the more normal DOUBLE_NONZERO style check (this applies for material properties like lambda which are usually manually set by input, not the fixed check eigen decomp)

Is the CAPILLARY change likely to see use in 3D, I'm okay with it is currently as it doesn't change the current behavior.
I know we don't actually calculate tangents in 3D so I guess we would need to start doing that and then make a backwards incompatible change to CAPILLARY_BC to get the 3rd float.

rbsecor:
I have no plans to use the CAPILLARY shear stress term in 3D, and hence, did not try to add the required additional float which would have required multiple other modifications.  But I thought I could lay some groundwork for its eventual inclusion.

And I thought we did compute tangents in 3D for any rotated condition?  That’s not the case?

Robert

wortiz:
> And I thought we did compute tangents in 3D for any rotated condition?  That’s not the case?

We do for the rotation but do not for the surface tangents at Gauss points in bc_integ.c (I believe we don't in collocated either). I don't think we have any BC's that use surface tangents in 3D that aren't manually specified tangents as far as I know

PR #306, opened 2020-08-05
Add regularized Saramito model and fix bug in 3D computation of deviatoric stress norm

jtmcconnell:
This merge request adds capability to run a regularized Saramito model for EVP flows. A bug present in the computation of the deviatoric stress tensor norm for 3D cases has also been fixed.

### DIFF
--- a/include/mm_fill_stress.h
+++ b/include/mm_fill_stress.h
@@ -203,8 +203,9 @@ compute_d_exp_s_ds(dbl [DIM][DIM],                   //s - stress
                    dbl [DIM][DIM][DIM][DIM]);        // d_exp_s_ds
 
 dbl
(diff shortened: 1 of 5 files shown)


wortiz:
.runguts

wortiz:
Could you add a pull request for documentation as well

https://github.com/goma/user-manual/blob/master/material_file/mechanical_and_constitutive/polymer_constitutive_equation.rst

As well as two cards for the polymer yield stress and polymer yield exponent

https://raw.githubusercontent.com/goma/user-manual/master/material_file/mechanical_and_constitutive.rst

They don't have to be super detailed but just so we know what we should be using

Math is supported as well if you want to add equations

https://www.sphinx-doc.org/en/1.0/ext/math.html

If you need more info about how to add them let me know

Issue #207, opened 2018-10-30
LSA - eigenvectors not added to most recent solution


DrBobF:
When performing linear stability analysis (LSA), the resulting eigenvectors (displacements) are not added to the most recent solution (deformed mesh) but instead are added to the FEM file.  This limitation can be overcome if the LSA is performed only for a single base flow by annealing the mesh and using it as the FEM file and then doing LSA.  But this is not typically the case; one is usually performing a continuation in a parameter (or several) and at each solution along the path performing the LSA.

PR #308, opened 2020-08-11
Anneal and un-anneal mesh automatically prior and after eigenvalue ca…

tjiptowi:
…lculation - ARPACK only


wortiz:
Does this fix #207?

src/ac_loca_interface.c
```diff
@@ -3423,7 +3426,20 @@ void solution_output_conwrap(int num_soln_flag,
       i_print = TRUE;
       step_print += cont->print_freq;
     }
-          
+
```
wortiz, line 3429: Maybe it would be better to put the allocations and frees in a displacement_somewhere if like the call so we don't allocate unnecessarily

src/rf_solve.c
```diff
@@ -3960,6 +3960,218 @@ anneal_mesh_with_external_field(const Exo_DB *exo)
...
+   */
+
+  if ( dim > 0 )
+    {
+     dcopy1( num_nodes, new_coord[0], exo->x_coord);
+    }
+
+  if ( dim > 1 )
+    {
+     dcopy1( num_nodes, new_coord[0], exo->x_coord);
```
wortiz, line 4075: Should this be new_coord[1] and exo->y_coord

src/rf_solve.c
```diff
@@ -3960,6 +3960,218 @@ anneal_mesh_with_external_field(const Exo_DB *exo)
...
+    }
+
+  if ( dim > 2 )
+    {
+     dcopy1( num_nodes, new_coord[0], exo->x_coord);
```
wortiz, line 4080: Similarly new_coord[2] and exo->z_coord

wortiz: I think you could do this directly instead of having a new_coord above right? Cause you already saved the coordinates beforehand 

so instead of

    new_coord[0][gnn] += saved_displacement[0][gnn];

You could do

    exo->x_coord[gnn] += saved_displacement[0][gnn];

PR #309, opened 2020-08-12
Extend list of continuation parameters in Goma

jtmcconnell:
This merge request will add polymer yield stress and yield exponent to the list of parameters over which continuation problems may be run.


wortiz:
.runguts

wortiz:
.runguts

PR #317, opened 2020-11-23
Time-Harmonic Maxwell's electric field equations and far-field BC

acochrane:
Includes term for conversion to heat in energy equation

Lots of potential "space junk" in mm_fill_em.c

-----------------------------------------------
Best - Curl of curl of E-field equation
     - Solve for single complex vector field
-----------------------------------------------
(EQ) assemble_ewave_curlcurl
(BC) apply_ewave_curlcurl_farfield_vec
-----------------------------------------------

Alternative - Laplacian of E-field
-----------------------------------------------
(EQ) assemble_ewave_laplacian
(BC) no integrated BCs implemented
-----------------------------------------------

Alternative - Curl of E and Curl of H
            - Solve for E and H complex vector
              fields simultaneously
-----------------------------------------------
(EQ) assemble_emwave
(BC) apply_em_*
-----------------------------------------------

Co-authored-by: Weston Ortiz <weston@wortiz.com>


acochrane:
.runguts

acochrane:
.runguts

src/mm_numjac.c
```diff
@@ -556,7 +556,7 @@ numerical_jacobian_compute_stress(struct Aztec_Linear_Solver_System *ams,
 	/*needed for saturation hyst. func. */
 	PRS_mat_ielem = ielem - exo->eb_ptr[ebn];
 
-	matrix_fill_stress(ams, x_1, resid_vector_1,
+	matrix_fill(ams, x_1, resid_vector_1,
```
wortiz, line 559: Is this still needed, do we want to keep the two equation form?

acochrane: I thought it was matrix_fill_stress so that numerical Jacobians would be computed for the two equation form.
Since the single equation form is the current performer, I think I set it back to matrix_fill so that no numerical Jacobians would be accidentally added in.

I could be mistaken about why this line changed.

wortiz: I think you had difficulties getting matrix_fill_stress to work so you changed it to matrix_fill 

It should be matrix fill stress, Robert just ran into an edge case where it is required to be different so we should make sure to have it `matrix_fill_stress` before merging

PR #323, opened 2021-02-05
Droplet evaporation models for aerosols

rbsecor:
Added some models for coarse-graining evaporation and diffusion from droplets in aerosol flows.  DROP_EVAP is a species source model, residence time field model for following relative droplet radius, and a heat source model.  Residence time field parameters such as diffusion constant and SUPG extent are now connected to the input file.

Some modification of DENSITY_IDEAL_GAS, more to be done here.  Also added a species source post-processing field to display species generation term


src/mm_input_mp.c
```diff
@@ -7670,7 +7689,7 @@ ECHO("\n----Acoustic Properties\n", echo_file);
    */
   rewind(imp);
 
-  if (read_bc_mp !=-1)
+  if (1 || read_bc_mp !=-1)
```
wortiz, line 7692: This looks like debugging left in by accident. Or is there something wrong with logic that needed this?

src/mm_post_proc.c
```diff
@@ -290,6 +290,7 @@ int Q_FCN = -1;
 int POYNTING_VECTORS = -1;   	/* conduction flux vectors*/
 int SARAMITO_YIELD = -1;
 int STRESS_NORM = -1;
+int SPECIES_SOURCES = -1;   	/* continuous species sources*/
```
wortiz, line 293: SPECIES_SOURCES needs to be added to dp_vif as well

wortiz (commented):
I still need to look through it a bit but spotted a couple things, left in separate code comments.

src/mm_input_mp.c
rbsecor: Oops, I think you're right.  Let me think of a better way of enabling that section - Thanks.

src/mm_fill_terms.c
```diff
@@ -34806,14 +34879,15 @@ assemble_poynting(double time,	/* present time value */
 	      if ( extended_dof && !xfem_active ) continue;
             }
 	  phi_i = bf[eqn]->phi[i];
+	  wt_func = (1.-mp->Rst_func_supg)*phi_i;
```
wortiz, line 34882: I don't think you want 1-SUPG_SCALE here.

Typically we put the scale only on the upwinding part effectively scaling tau, a common choice is 1.0 so with this I think you lose your phi_i portion

Some of Goma has odd tau choices as well. Typically you choose one like equation 4 here https://www.wias-berlin.de/people/john/ELECTRONIC_PAPERS/JK07.CMAME.pdf where as elem size -> 0 the upwinding disappears and you scale with 1/mag(v). The newer species one works well if you are having stabilization issues, I've also found SUPG works better when applied across all terms you can (avoiding the usual only on advection which ends up overly diffusive).

src/mm_fill_terms.c
```diff
@@ -34865,14 +34939,12 @@ assemble_poynting(double time,	/* present time value */
...
```
wortiz, line 34942: I think this ends up overwriting the previous comment. But it still looks like the one you probably don't want

src/mm_fill_terms.c
rbsecor: This was intentional.  These residence time fields work best with only the gradient basis function term which 1-supg allows for supg = 1.  i.e., just the Petrov piece and no Galerkin in Petrov-Galerkin.  I guess I could change it to 2-supg so that supg=1 gives the Petrov-Galerkin term similar to other parts of Goma, but with supg=2 we get only the Petrov piece.

wortiz: Okay, I wanted to be sure that was the intent. I think we can leave it for now.

Issue #324, opened 2021-02-19
Parallel read exoII file errors with Element variables (P_E1, etc.)


rbsecor:
**Example:**
Initial guess read from "cont.exoII.2.0" ...(last soln in file)
Nodal variable VX found in exoII database - reading.
Nodal variable VY found in exoII database - reading.
Nodal variable T found in exoII database - reading.
P_1 ERROR EXIT:rf_util.c:2692: ex_get_var element
P_1 ERROR EXIT:rf_util.c:2692: ex_get_var element
Nodal variable CK_0 found in exoII database - reading.
P_1 ERROR EXIT:rf_util.c:2692: ex_get_var element
Nodal variable DMX found in exoII database - reading.
Nodal variable DMY found in exoII database - reading.
Element variable P_E1 for material 1 found in exoII database - reading.
P_0 ERROR EXIT:rf_util.c:2692: ex_get_var element
Element variable P_E2 for material 1 found in exoII database - reading.
P_0 ERROR EXIT:rf_util.c:2692: ex_get_var element
Element variable P_E3 for material 1 found in exoII database - reading.
P_0 ERROR EXIT:rf_util.c:2692: ex_get_var element

wortiz:
What kind of problem is this?

We tested Continuation/Steady/Transient/LSA for sure.

Is this Hunting?

rbsecor:
Steady continuation:
e.g.,
Number of Materials = 2
 MAT =   liquid   1
 
  Coordinate System = CYLINDRICAL
  Element Mapping   = isoparametric
  Mesh Motion = ARBITRARY
  Number of bulk species = 1
  Default Material Species Type = SPECIES_CONCENTRATION
 
        Number of EQ  = 7
        EQ = mesh1      Q2      D1      Q2    	 
        EQ = mesh2      Q2      D2      Q2      
        EQ = momentum1  Q2      U1      Q2    	    
        EQ = momentum2  Q2      U2      Q2          
        EQ = continuity P1      P       P1      
   	EQ = energy     Q2      T       Q2   
        EQ = species_bulk Q2    Y       Q2      1.  1.   1.   1.   1.   0.

rbsecor:
Parallel runs fine when not reading an initial guess from cont.exoII.  It's the next run where cont.exoII is broken and read in that the error occurs.

wortiz:
Ok thanks I'll take a look

wortiz:
[...]

PR #325, opened 2021-02-22
Don't write nodal variables after free

wortiz:
I think this will address #324 @rbsecor


wortiz:
Note standalone brkfix also updated. Should be usable until this is merged

PR #327, opened 2021-03-19
Droplet Evaporation Part 2

rbsecor:
In this addition, I managed to make a "medium" dive through mm_sol_nonlinear.c and convergence criteria.  Some logical errors were found, but unfortunately virtually every test suite problem fails due to convergence and rate differences.  I think we can all agree that converged should be TRUE instead of FALSE, but all may not like the output format.

  ToD    itn   L_oo    L_1     L_2     L_oo    L_1     L_2   lis  asm/slv (sec)  L2_Order/Rates
-------- --- ------- ------- ------- ------- ------- ------- --- --------------- -------------
13:41:59 [0] 9.3e-08 7.6e-06 2.0e-07 5.8e-03 2.0e+02 1.2e+00  1  5.6e+01/2.0e+02 0.00/0.00 0.00/0.00
13:46:16 [1] 3.4e-07 2.0e-05 7.8e-07 4.6e-03 1.4e+02 9.0e-01  1  5.5e+01/2.0e+02 0.97/-0.60 1.01/0.12
13:50:31 [2] 5.0e-07 1.9e-05 1.1e-06 3.2e-03 1.0e+02 6.4e-01  1  5.6e+01/2.0e+02 0.26/0.07 1.22/0.16
13:54:44 [3] 4.8e-07 1.6e-05 1.0e-06 1.0e-03 3.2e+01 2.0e-01  1  5.6e+01/2.0e+02 -0.19/0.12 3.42/0.68
13:59:00 [4] 2.8e-07 9.5e-06 6.0e-07 4.9e-04 1.7e+01 9.3e-02  1  5.5e+01/1.9e+02 8.08/0.34 0.68/0.26
-------- --- ------- ------- ------- ------- ------- ------- --- --------------- -------------
scaled solution norms   2.440e+04  2.415e+02  1.706e+03 

It seems important to separate out convergence order from convergence rate.  i.e., convergence order can be nominal while we are diverging rapidly.  I currently have 3 point calculation enabled but we could go back to 2 point rules. Thoughts?


wortiz:
Right now one advantage of the output format is that it comes out to the standard 80 characters so it fits in default terminals (at least for direct solves + low iterations)

I have to look at it more to give any real opinion. Maybe @tjiptowi will have a quicker opinion.

We don't usually look at rates explicitly so having it displayed might be nice.

Things I don't really look at in the current output are the time and L_1 norms so maybe those could be used to trim output if others are similar in their thinking.

rbsecor:
We could make some parts of the output optional, maybe controllable through the input file.  I guess we would always want it trimmed down to 80 characters.
Also, looks like I have a SERR test failure to look at...

src/mm_fill_species.c
```diff
@@ -11029,9 +11060,27 @@ get_continuous_species_terms(struct Species_Conservation_Terms *st,
...
+	   }
+	else
+	   { dsdT = s*(-1./fv->T) + tmp*rad_ratio *
+		(dpsatdt[w]/pres_conv*x_mole[w]/(1-activity[w])
+		-dy_dT/(1-y_mole[w]));
+	   }
+	dsdrst = tmp*log((1-y_mole[w])/(1-activity[w])) + tmp*rad_ratio*dact_dr/(1.-activity[w]);
+
+#if 1
+if(isnan(s))fprintf(stderr,"dropnan %g %g %g %g %g %g\n",s,tmp,y_mole[w],activity[w],A,x_mole[w]);
```
wortiz, line 11078: If you are looking for NaNs and have a problem that exhibits them, compiling with the FP_EXCEPT define: https://github.com/goma/goma/blob/6a4659b310f6f0a625aebb579ce69f1a4d97fdab/settings.mk-example#L86

Will cause gdb and other debuggers to break on the first action that creates a NaN

wortiz:
It looks like porous_deform is failing, I'm not sure if it is something you did or another bug being brought up that was hidden.

> We could make some parts of the output optional, maybe controllable through the input file.

I think we probably would want to clean up the printing of the Order/Rates to be included by default. But making parts optional based on input cards sounds pretty good.

rbsecor:
Yes, I vaguely recall inadvertently running into some porous code.  I'll see if I can remember where that was and work on the output a bit.

rbsecor:
With the last push, input file control of the solver output format is introduced via bit-wise column selection - "Solver Output Format Bitmap = integer.  1023 for traditional format, 4095 is the maximum (2^12-1).  I guess bigger numbers would work but only the last 11 bits are active.

wortiz:
Okay I'll try to demo it, I do like the idea of optional output.

rbsecor:
Adjusted the output format a bit and added a RESTIME_GRADSIC_BC bc which I am testing out.  Some cleanup preparing for multiple solvents in the droplet evaporation model.

src/bc_integ.c
```diff
@@ -1744,6 +1744,12 @@ apply_integrated_bc(double x[],           /* Solution vector for the current pro
 	  qnobc_surf (func, d_func, time_intermediate);
 	  break;
 
+	case RESTIME_GRADSIC_BC:
+	  func[0] = -bc->BC_Data_Float[0];
```
wortiz, line 1748: I'm guessing you want a `break;` here since it is a strong condition but if you want it to fall through put an explicit comment:

    /* fall through */

like here https://github.com/goma/goma/blob/b4ce3c28f7b4a1ae9a3c8772bbb908d0501a51b0/src/mm_fill_ls.c#L8256-L8259

I think guts check should report a warning generated here but if not I need to update the testing compiler

wortiz:
I looked at this some more I'll leave a few comments and then bring up the pull request during the Goma local meeting this week.

src/mm_matrl.c
```diff
@@ -721,13 +723,13 @@ calc_density(MATRL_PROP_STRUCT *matrl, int doJac,
 			       matID, num_species-1, 0.0, rho);
 	    break;
 	case SPECIES_CONCENTRATION:
-	    for (w = 0; w < num_species - 1; w++) {
-	      tmp =  mw[w] - mw[num_species - 1];
+	    for (w = 0; w < num_species_eqn; w++) {
+	      tmp =  mw[w] - mw_last;
```
wortiz, line 727: Looks like mw_last is uninitialized when the if is taken on line 669

src/mm_std_models.c
```diff
@@ -2431,9 +2431,11 @@ Diffusivity (void)
       ********************************************************************/
 {
   int i, j, w, var, err= 0;
+  double T=298.0, P=101.325;;
   for (w = 0; w < pd->Num_Species; w++) {
     switch (mp->DiffusivityModel[w]) {
 
+  char err_msg[MAX_CHAR_IN_INPUT];
```
wortiz, line 2438: err_msg is unused

src/mm_sol_nonlinear.c
```diff
@@ -1291,7 +1335,80 @@ EH(-1,"version not compiled with frontal solver");
...
 	   
-	  if( Linear_Solver != FRONT && *converged ) goto skip_solve;
+      if( (Linear_Solver != FRONT) && *converged && enable_skip_solve) 
+        {
+/*  If we're going to skip the Solve step, compute and save some useful 
+    information first...*/
+
+         Resid_Norm_stack[0] = Resid_Norm_stack[1];
+         Resid_Norm_stack[1] = Resid_Norm_stack[2];
+         Resid_Norm_stack[2] = Norm[0][2]/BIG_PENALTY;
```
wortiz, line 1345: Why the big penalty on the norm?  Without it the order looks more like what I would expect.

This is also changing to L2 which I think should be near the same to L1 so that's probably fine.

For example, for:

    11:17:13 [0] 1.3e-03 1.7e-02 4.1e-03 1.3e-03 1.8e-01 1.4e-02  1  3.0e-02/8.0e-02 
    11:17:13 [1] 4.3e-14 1.5e-12 1.5e-13 5.4e-13 6.1e-11 4.9e-12  1  2.0e-02/3.0e-02

Which looks like it should report an estimate that is quadratic convergence. 

For the BIG_PENALTY version we get:

    CONVRATE     rel diff:  6.6759460e+00 ~  1.7241648e+00 = 7.41735e-01 (FAILED)

Which is an estimate below quadratic order, but without big penalty it looks correct:

    CONVRATE     rel diff:  6.6759460e+00 ~  5.3590304e+00 = 1.97263e-01 (FAILED)

src/mm_sol_nonlinear.c
```diff
@@ -1943,13 +2139,9 @@ EH(-1,"version not compiled with frontal solver");
 	  convergence_rate_tolerance != 0.) {
 	
 	/* Compute convergence level and compare with tolerance */
-	Norm_below_tolerance = (Norm_new < modified_newt_norm_tol);
+	Norm_below_tolerance = (Resid_Norm_stack[2] < modified_newt_norm_tol);
```
wortiz, line 2142: I think this is changing from L_1 to L_2 but I think the manual specifies L_2 so probably okay

src/mm_sol_nonlinear.c
```diff
@@ -1943,13 +2139,9 @@ EH(-1,"version not compiled with frontal solver");
...
 	
-	/* Compute rate_of_convergence.  Really what we want
-	   here is dln(norm)/dnorm   FIGURE IT OUT! */
-	/* Rate_above_tolerance = (pow(Norm_old, 1.8) > Norm_new); */
-	Rate_above_tolerance = ((log10(Norm_new)/log10(Norm_old)) >
-				convergence_rate_tolerance);
+	Rate_above_tolerance = (Conv_rate > convergence_rate_tolerance);
```
wortiz, line 2144: I think we want Conv_order to be consistent here, manual equations should probably be updated.

Manual also specifies L_1 and this is L_2

https://docs.gomafem.com/problem_description_file/solver_specifications.html#modified-newton-tolerance

rbsecor:
Thanks.  I'll go through and cleanup a number of these items.  The reason for the BIG_PENALTY scaling was to avoid negative convergence orders which occurs when a Norm crosses from > 1 to < 1.  Since the convergence rates and orders are relative the magnitude is irrelevant to the final answer.  A main reason for differences from old results is the use of 3-point rules instead of 2-point, i.e. using last 3 results instead of 2.  3-point rules should be more accurate (as well as "more correct" for the convergence order, i.e. no assumption on lambda) but may be less stable.  We can take a look and decide to revert back to 2-point rules - there's a couple #if blocks to switch back.  I was waiting to see how the results panned out on a few problems to decide.

I assumed L2 was always used as a basis and didn't look at the manual.  It shouldn't be that different between the two but we should be consistent.  I'll take another look.

rbsecor:
A couple more comments on convergence rate/order:
1)  What is currently being written to the exoII files as CONVRATE is actually the convergence order.  Imo this provides limited utility as the order doesn't indicate whether an iteration sequence is converging or diverging, for example.  I would surmise a more useful indicator would the convergence rate, i.e. negative slope of semilog plot of error measure vs. iteration number.
2) Previously, some code blocks were out of order causing the reported convergence rate (actually the convergence order) to lag by 1 iteration.  As a result, there will be test suite differences even if the previous formula for convergence rate (order) were to be re-instated.

wortiz:
> What is currently being written to the exoII files as CONVRATE is actually the convergence order. Imo this provides limited utility as the order doesn't indicate whether an iteration sequence is converging or diverging, for example. I would surmise a more useful indicator would the convergence rate, i.e. negative slope of semilog plot of error measure vs. iteration number.

I think we would be okay changing this. I think right now the only thing really using this is GUTS and we already require a rebaseline. I think it really only is to help us notice if we change something from quadratic to linear that the other tests don't catch.

rbsecor:
A continuation of developments - 2-1/2 D CARTESIAN updates for viscous and VE calculations which involved moving wim calcs to the global WIM.  Also fixed issues with velotheta test problem (mostly) and others related to addition of external shear stress on CAPILLARY-type boundary conditions

rbsecor:
Mostly TALE fixes; (1) corrected wall velocity calc for VELO_THETA type wetting conditions, (2) Enabled the "rolling motion" wetting condition for FSI problems, e.g. use with VELO_TANGENT_SOLID and VELO_SLIP_SOLID.

src/mm_sol_nonlinear.c
```diff
@@ -1924,12 +2058,91 @@ EH(-1,"version not compiled with frontal solver");
...
+	 if(inewton <= 1)
+	    {
+	     AConv_order = log10(AC_Resid_Norm_stack[2])/log10(AC_Resid_Norm_stack[1]);
+	     ASoln_order = log10(AC_Soln_Norm_stack[2])/log10(AC_Soln_Norm_stack[1]);
+	     AConv_rate = log10(AC_Resid_Norm_stack[1])-log10(AC_Resid_Norm_stack[2]);
+	     ASoln_rate = log10(AC_Soln_Norm_stack[1])-log10(AC_Soln_Norm_stack[2]);
+	    }
+	 else
+	    {
+	     Conv_order = log10(AC_Resid_Norm_stack[2]/AC_Resid_Norm_stack[1])
```
wortiz, line 2129: I think these should be AConv_order ... etc

Also AC's seem to be more likely to have 0 residual norms so might be worthwhile to add an epsilon to the denominator for at least residuals.

Right now this is showing up as a NaN in the exodus output in volconst due to the wrong name and division by zero

wortiz:
I think I'm mostly through this.

There is the AConv_order and NaN issue I just pointed out.

Another issue is for velotheta you added a required float for CAP_REPULSE so that problem is failing. it looked like maybe you changed order as well?

Idk if we want to keep old behavior or move to your new one but if you could let me know how to update the problem and docs that would be good.

Let me know if you need access to guts again

wortiz:
.runguts

wortiz:
.runguts

rbsecor:
OK, I’ll take a look.

 

Robert

wortiz:
.runguts

rbsecor:
I’ve been wondering if I’m connecting to the correct guts repository.  Is 129.24.4.218 the correct machine?

 

Thanks,

Robert

rbsecor:
p.s. I’m aware of the extra required float – I think it maybe it affected other problems besides velotheta (I’ll have to look again).  I wasn’t aware of changing order – do you mean Q2 vs Q1 etc.?

wortiz:
I think I was thinking of these when I said change in order:

https://github.com/goma/goma/pull/327/files#diff-703d8eb1a9a3f82e56b0dfe4b1a7a9b2a5b265a71a0d693880a39cce381eacbaR1005-R1007

That is the correct ip for guts I sent an email to followup separately

wortiz:
That link takes a long time to load, here is the relevant diff you had just put a float at the beginning I think so the data floats went +1

	    if (bc->BC_Name == CAP_REPULSE_BC) {
	      apply_repulsion(cfunc, d_cfunc, bc->BC_Data_Float[2],
			      bc->BC_Data_Float[3], bc->BC_Data_Float[4],
			      bc->BC_Data_Float[5], bc->BC_Data_Float[6],
-----------------------------
	      apply_repulsion(cfunc, d_cfunc, bc->BC_Data_Float[3],
			      bc->BC_Data_Float[4], bc->BC_Data_Float[5],
			      bc->BC_Data_Float[6], bc->BC_Data_Float[7],
			      elem_side_bc, iconnect_ptr);
	    }

rbsecor:
Yes, the add float is the applied shear stress which naturally goes after the added normal stress (pressure).  So, all the other floats move down one.

 

Robert

rbsecor:
If I did this correctly, I added an eps to the residual norms to avoid Nans.  As long as we're rebaselining the test suite, I added a global variable CONVORDER (the order of convergence).  Before Goma was reporting the convergence order as a convergence rate.  Now the convergence order is reported as CONVORDER and the actual convergence rate is reported as CONVRATE.  My assumption would be that the convergence rate is more useful.

Definitions: Convergence rate = - semilog slope of residual error vs. iteration number
                   Convergence order = exponent in error_(i+1) = K (error_i)^p
Note you would get a convergence order of 2 for an iteration sequence that is rapidly diverging...

Also, TALE and Convective Lagrangian Velocity updates in this version.

wortiz:
.runguts

wortiz:
I think guts is updated except for velotheta

I might need to update Jenkins build / some tolerance settings for exodiff for the suitesparse rebaseline.

Everything was rebaselined using Ubuntu 20.04 gcc 9.3.0 with the latest build script with a new mpi and suitesparse I think suitesparse should be the only thing that would change exodiff if this latest run has more than 1 failure.

rbsecor:
To update velotheta files, do I just push files to the guts master branch?  I assume there is just one branch – not the multiple we have for the goma github.

 

The addition of a global variable broke the guts exodiffs, I believe.  So maybe another rebaselining is in order.

 

Robert

wortiz:
Yes you can update master just don't do any destructive git commands like force push etc.., I've updated everything with CONVORDER now except for velotheta

wortiz:
.runguts

wortiz:
.runguts

PR #331, opened 2021-06-16
remove distfact requirement for reading exodus files

wortiz:


wortiz:
This change actually requires more substantial work...closing for now

PR #340, opened 2021-12-16
Variable Stats

rbsecor:
Check in of the optional variable statistics computation along with some other miscellaneous stuff.  Some fixes for combo shell problems, etc.

Solver Output Format Bitmap	 = {2^12-1}
Output Variable Statistics	 = yes


rbsecor:
Do we know the offending guts problems?  I could guess maybe the surf rheo ones...

wortiz:
plumb_bob_2 and plumb_bob_5_noMesh

plumb_bob_2 is just L_inf norm slightly different, plumb_bob_5_noMesh has different convergence and some exodus differences that need to be looked at

src/mm_fill_util.c
```diff
@@ -339,7 +338,7 @@ beer_belly(void)
    * computed above, so use this block with caution.
    */
   if(elem_shape == SHELL
-     || elem_shape == TRISHELL
+     || elem_shape == TRISHELL || elem_shape == LINE_SEGMENT
```
wortiz, line 341: This is why guts is failing, I think pd->Num_Dim will be 1 for a BAR3 that isn't a shell and then you will end up with the wrong Jacobian. If I understand the code correctly

rbsecor: I'll give it a try.

src/mm_fill_aux.c
```diff
@@ -1428,6 +1428,9 @@ surface_determinant_and_normal(
           EH(-1, "Incorrect side for TRIANGLE");
         }
       break;
+    case LINE_SEGMENT:
```
wortiz, line 1431: Is this an extra case that isn't needed, similar to the beer_belly BAR3 comment, or are we maybe missing something for LINE shells.  If so maybe that beer_belly and here needs a SHELL1D or something shape to distinguish it from non shell LINE_SEGMENTS

src/rf_solve.c
```diff
@@ -4062,5 +3945,104 @@ npost[i] = Export_XP_ID[i]; }
...
+		}
+	  }
+	}
+     }
+/*
+ * Find global quantities
+ */
+#ifdef PARALLEL
+  MPI_Allreduce( (void*)max, (void*)max_buf, MAX_VARIABLE_TYPES,
+                MPI_DOUBLE, MPI_MAX, MPI_COMM_WORLD);
```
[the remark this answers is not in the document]
rbsecor: I think I copied this from mm_flux.c which seemed like a similar function.  It sounds like both of these should MPI_Reduce.  Do we need to check all the MPI_Allreduces?

src/mm_fill_aux.c
rbsecor: It's possible this is superfluous.  One of these additions provided for an improved solution, but that could have been do to other changes.  I'll try to double-check.  I wasn't aware we had the possibility of 1D elements that were not shells...

src/rf_solve.c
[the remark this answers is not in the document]
wortiz: I think the difference for evaluate_flux is that we use it across all processors (for augmenting conditions / level set mass conservation) whereas this is only needed on the processor printing the statistics.

The only difference between Reduce and Allreduce is that Allreduce the resulting data from the reduction is available on all processors whereas reduce it is only available on the root.

Looking through quickly it looks like current Allreduce result is used on all/many processors for most of the clals I looked at.

src/mm_fill_aux.c
wortiz: You might be right here, maybe the `beer_belly` issue is more related to [@acochrane's comment](https://github.com/goma/goma/blob/b4a62878dab0b18fc9bc5119da7d45d48b3f6a5e/src/mm_fill_util.c#L333-L339) regarding the Jacobian adjustment not necessarily working for certain elements.

I do think this is probably superfluous as it should be covered by `shell_determinant_and_normal` unless you have a use case for it.

PR #341, opened 2021-12-21
Update SH_FLUID_STRESS implementation

tjiptowi:


src/bc_colloc.c
```diff
@@ -558,14 +561,75 @@ xsurf[2] = BC_Types[icount].BC_Data_Float[BC_Types[icount].max_DFlt+3];
...
+
+	         func = 0.; /* this boundary condition rearranges values already in res and jac,
 			    * and does not add anything into the residual */
-	      break;
+	         break;
+                }
+              else
+                {
+                 break;
+                }
```
wortiz, line 632:
```suggestion
                }
                break;
```

src/mm_fill_shell.c
```diff
@@ -4306,9 +4296,9 @@ put_fluid_stress_on_shell(int id, /* local element node number for the
       ieqn_shell = R_SHELL_CURVATURE;
       ieqn_mom1   = R_MOMENTUM1;
       ieqn_mom2   = R_MOMENTUM2;
-      id_dofmom1 = ei->ln_to_dof[ieqn_mom1][id];
-      id_dofmom2 = ei->ln_to_dof[ieqn_mom2][id];
-      id_dofshell = id_shell; 
+      id_dofmom1 = id;
+      id_dofmom2 = id;
```
wortiz, line 4300: Shouldn't `id_dofmom2` correspond to VY here

src/mm_fill_shell.c
```diff
@@ -4306,9 +4296,9 @@ put_fluid_stress_on_shell(int id, /* local element node number for the
...
+      id_dofshell = id_shell;
```
wortiz, line 4301: I think these `id_shell` are local node numbers rather than `DOF` indices if I'm following the logic correctly so I think you are going to end up with wrong residual contributions (unless it works out by happenstance).

tjiptowi: Correct. My understanding is that local node number and DOF is one of the same, except for P1 basis functions

src/mm_fill_shell.c
tjiptowi: Correct. I am just using the shortcut that local node number and DOF is the same. The way we use Lec->R and Lec->J is pretty much with local node number instead of DOF.

src/mm_fill_shell.c
wortiz: I think there are a few, P0, P1, Q1_D, Q2_D, xfem and ghost types.

I'm not sure if any of those would be relevant here.

wortiz:
I think there is a bug in the implementation `slot_shell_q1` and `slot_shell_q2` have very different final results there appears to be no mesh deformation of the shell.

src/bc_colloc.c
```diff
@@ -558,14 +561,75 @@ xsurf[2] = BC_Types[icount].BC_Data_Float[BC_Types[icount].max_DFlt+3];
...
+
+              for (i=0; i<DIM; i++)
+                 {
+                  xi2[i] = 0.0;
+                 }
+
+              dim1 = elem_info(NDIM, Elem_Type(exo, el1));  /* Local element */
+              dim2 = elem_info(NDIM, Elem_Type(exo, el2));  /* Remote element */
+              ddim = dim1 - dim2;
+              if (ddim == -1 && id != -1)
```
wortiz, line 579: We only seem to get in here on the bulk side so this always fails, maybe this is a limitation with bulk/shell BCs and sidesets are only on the bulk?

tjiptowi: The version before only applies SH_FLUID_STRESS when we are on the bulk side and that causes problem with Dirichlet boundary conditions for shell equations with bulk element next to it.

wortiz: `slot_shell_q1` and `slot_shell_q2` don't get past this if as ddim is always -1, I'm not sure if we need to redo those mesh files so the sidesets are such that we will get in there on the shell element or if this is just a limitation in the current Goma BC logic

PR #363, opened 2022-03-21
Hele-shaw-model_for_BCY_using_formulation_by_Bonn

bikash029:


wortiz:
I think we will have to look at this in a bit more detail, an initial merge is available at:

[bonn_lub branch /goma/goma/tree/bonn_lub](/goma/goma/tree/bonn_lub)

Issue #346, opened 2022-02-25
GOMA 6 crashes if number of unknowns is bigger than ~560000


AndrewUshakov:
Hi,

I have a test case where solution algorithm is GMRES, number of unknowns = 562432 and  number of matrix nonzeroes = 58436224. If matrix format is MSR, stacktrace is:

```[ubuntu:03973] *** Process received signal ***
[ubuntu:03973] Signal: Segmentation fault (11)
[ubuntu:03973] Signal code:  (128)
[ubuntu:03973] Failing at address: (nil)
[ubuntu:03973] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0)[0x7ffff77c23c0]
[ubuntu:03973] [ 1] /home/andrew/sandia/2021/goma/cmake-build-debug/goma(AZ_MSR_matvec_mult+0x181)[0x5555564a3261]
[ubuntu:03973] [ 2] /home/andrew/sandia/2021/goma/cmake-build-debug/goma(AZ_compute_residual+0x27)[0x555556499f47]
[ubuntu:03973] [ 3] /home/andrew/sandia/2021/goma/cmake-build-debug/goma(AZ_scale_true_residual+0x88)[0x5555564a1278]
[ubuntu:03973] [ 4] /home/andrew/sandia/2021/goma/cmake-build-debug/goma(AZ_pgmres+0x1352)[0x5555564bec52]
[ubuntu:03973] [ 5] /home/andrew/sandia/2021/goma/cmake-build-debug/goma(AZ_oldsolve+0x49c)[0x5555564acf8c]
[ubuntu:03973] [ 6] /home/andrew/sandia/2021/goma/cmake-build-debug/goma(AZ_solve+0x1d8)[0x5555564adbf8]
[ubuntu:03973] [ 7] /home/andrew/sandia/2021/goma/cmake-build-debug/goma(solve_nonlinear_problem+0x2f2b)[0x555555e01f9e]
[ubuntu:03973] [ 8] /home/andrew/sandia/2021/goma/cmake-build-debug/goma(solve_problem+0x1666)[0x555555e41aba]
[ubuntu:03973] [ 9] /home/andrew/sandia/2021/goma/cmake-build-debug/goma(main+0xe55)[0x555555bcb371]
[ubuntu:03973] [10] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7ffff75e20b3]
[...]

PR #370, opened 2022-04-19
Fix issue #346 (GOMA 6 crashes if number of unknowns is bigger than ~560000)

AndrewUshakov:
Patch files az_aztec.h and az_util.c from the aztecoo package in the Trilinos 13.2.0 to the versions, suggested in the pull request https://github.com/trilinos/Trilinos/pull/10436, which is still not reviewed. See also the issue https://github.com/trilinos/Trilinos/issues/10334.

### DIFF
--- a/scripts/build-goma-dependencies.sh
+++ b/scripts/build-goma-dependencies.sh
@@ -2741,6 +2741,128 @@ else
 fi
 
 #make trilinos
+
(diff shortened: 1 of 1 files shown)


wortiz:
.runguts

wortiz:
I'll have to try this patch manually, but thanks for submitting it.

wortiz:
Hopefully Trilinos will merge this in next release.

PR #377, opened 2022-04-29
Goma7B User Mat Init

rbsecor:
Material initialization update and bugfixes ; Changes to remove compiler warnings
1. Added user_mat_init routine in user_pre.c for user material initialization
2. Constant vs. User material initialization are differentiated by "Initialize" vs. "User Initialize".
3. Filled out more shear-thinning, lubrication, stationary wall models - testing in progress


src/mm_shell_util.c
```diff
@@ -4658,7 +4671,7 @@ void calculate_lub_q_v(const int EQN, double time, double dt, double xi[DIM], co
       k_turb = 3.;
       dq_gradp = pre_delP = -CUBE(H) / (k_turb * mu) - beta_slip * SQUARE(H);
       q_mag = pre_delP * pgrad;
-      // dv_gradp = vpre_delP = pre_delP / H;
+      vpre_delP = pre_delP / H;
```
wortiz, line 4674: Sorry about that.

rbsecor: No problem.  I need to some "combing" over the lub_film part of lub_q_v -- not sure I understand exactly how the shear-thinning versions should look.

src/mm_fill_potential.c
```diff
@@ -381,44 +367,6 @@ int assemble_potential(double time, /* present time value */
...
-    if (pd->v[pg->imtrx][MASS_FRACTION]) {
-      for (w = 0; w < pd->Num_Species_Eqn; w++) {
-        var = MASS_FRACTION;
-        var_offset = MAX_VARIABLE_TYPES + w;
-        for (j = 0; j < ei[pg->imtrx]->dof[var]; j++) {
-          dhdC[w][j] = 0.0;
-        }
-      }
-    }
-
```
wortiz, line 421: Are you removing these because of the memset warnings from GCC with -O3?

Like you might see here: 

https://github.com/goma/goma/runs/6232827830?check_suite_focus=true#step:6:94

I feel like I looked for these before and there is a bug in GCC for these warnings.

rbsecor: Yes, I was getting compiler warnings to use memset.  I guess I didn't check on compiler options is use...

wortiz: I think everything should compile without warnings with -O2 -DNDEBUG, or -O2

I don't know which optimizations cause them but it seems with -O3 quite a few spurious warnings are generated.

With the current cmake setup if you build with:

-DCMAKE_BUILD_TYPE=RelWithDebInfo it will use -O2 -g I think which ends up getting nicer errors on GOMA_EH and is what all the tests are run against

-DCMAKE_BUILD_TYPE=Release I think all tests should pass but it uses the -O3 which gives some warnings, in the past it was breaking a couple test problems too.

wortiz:
Something is going on with `p_edge` but I'm getting it in #379 as well so I have to dig into it more.

wortiz:
.runguts

PR #385, opened 2022-05-11
Fix bug on element variable fixing, and running overloaded

wortiz:
Fixes a bug with `fix` when a mesh is broken enough that the element variables don't exist on all processors.


wortiz:
@BWagoner901 This should fix your element variable issue

PR #388, opened 2022-05-27
Lub wall

rbsecor:
Update of shear-thinning lubrication algorithms for stationary wall applications.  Viscosity models include power-law, Carreau, and Bingham styles.  The Carreau and Bingham varieties are done using numerical integration of the shearing-thining form of Reynolds Equation.  There seems to be some loss of quadratic convergence for the Bingham models for certain parameter values, which is still under investigation.  Both confined flow and film flow versions should work.

Also, some bugfixes and clean up of the Initialization bounds for variable initial conditions.


wortiz:
The failing problems are p_lub_pore and lub_shell, lub_shell doesn't look like a big issue, p_lub_pore looks like maybe something is wrong pretty big differences in LUB_VELO_*

rbsecor:
I don't know that I will find the issue with lub_shell.  I swapped each of changed files with goma-main versions and the exoII comparison I get for lub_shell is the same.  From my guts history, I see that something changed on about April 29, but am a bit bemused about what it could be.  Thoughts?

wortiz:
I think the last change for lub_shell was a rebaseline for the new compiler for Jenkins tests.

I think I need to add an absolute threshold for LUB_VELO_Z on that problem because relative to the rest of the problem it is very much 0 and this seems to crop up a few times, either changing compilers or it looks like this one was probably changing order of a few floating point operations. I'll take a look shortly.

wortiz:
.runguts

wortiz:
I think lub_shell is just a result from the change in order of operations in calculate_lub_q_v so should be okay to rebaseline. I should get to it today I'm having another odd problem with another guts problem on my machine that I need to take a look at first

PR #394, opened 2022-07-28
Added standard AC_ANGLE augmenting condition

rbsecor:
AC_ANGLE is an augmenting condition where the DCA at a node, typically DCL node, would be maintained at a constant value.  Typical usage might be to keep the DCA at some value slightly less than 180 degrees by moving a DCL fence.

Also some other mods, non-Newtonian lubrication shells, etc.  The User Initialize add-on is currently disabled as it's still buggy.


wortiz:
The failures are from LS problems foam, wedge_se, wedge_sg. I'm guessing its from your bugfix in mm_input_bc 

I'll have to look at the results, but I think the code looks like you fixed an issue.

wortiz:
.runguts

wortiz:
.runguts

wortiz:
.runguts

wortiz:
.runguts sorry for the spam I'm not sure why Jenkins is having so many issues on this one

wortiz:
.runguts

PR #396, opened 2022-08-28
Non-Newtonian Lubrication Update

rbsecor:
Next set of updates for non-Newtonian lubrication flow, particularly enabling LS problems whereby the  lubrication flow vector is modulated across the LS interface.  I.e. q_lub_air >> q_lub_liquid for the same pressure driving force, which should lead to small pressure variations in the "gas" phase similar to what might be expected from a low gas viscosity.

Also, added a GRAD_LUBP_NOBC bc (open flow condition) on the lubrication pressure.  As opposed to the NOBC on Navier-Stokes stress, pressure level does not appear in this NOBC.  Fixed up the LS_DCA post-processing flux Q2s (hopefully also Q1s).
Epsilon regularization also added to the lubrication BINGHAM models -- not sure if this is helpful yet, still testing.


wortiz:
.runguts

wortiz:
lub_shell seems particularly sensitive, I changed the tolerance on LUB_VELO_Z, it looks like differences are still small especially for a post processed variable.

Here is the typical epsilon regularization (eta here):

image

I think it looks a bit different than what you were doing if you have trouble with your way.

PR #408, opened 2022-10-25
October Lubrication Update

rbsecor:
- Added more viscosity integral options for shear-thinning liquids
- Attempted to enable lubrication free surface ALE methods which required id_side, T, and t calculation mods
- Added LUB_KINEMATIC bc which is still a work in progress
- Other miscellaneous code cleanup


src/mm_shell_util.c
```diff
@@ -3220,14 +3250,54 @@ void lubrication_shell_initialize(int *n_dof,        // Degrees of freedom
...
+      case -1:
+        T[0][0] = 1.;
+        T[1][1] = 1.;
+        break;
+      default:
+        GOMA_EH(GOMA_ERROR, "Incorrect side for SHELL");
+        break;
+      }
+      break;
+    }
```
wortiz, line 3300: I think the GUTS issues are coming from the change in calculating T here.

wortiz: And the other corresponding part in shell_determinant_and_normal

wortiz: It might be more the shell_determinant_and_normal since you checked this one unless I'm missing something else.

If you use the current main branch mm_shell_util.c the test suite passes.

I thought the first two side id's correspond to faces and then the remaining 4 are the usual quadrilateral sides but that doesn't seem to be the case in surface_determinant_and_normal so maybe we do something special to switch them from the exodus form

src/mm_shell_util.c
```diff
@@ -3992,6 +4068,21 @@ void calculate_lub_q_v(const int EQN, double time, double dt, double xi[DIM], co
...
+      q_mag =
+          ls_modulate_property(q_mag, q_mag2, ls->Length_Scale, (double)mp->mp2nd->viscositymask[0],
+                               (double)mp->mp2nd->viscositymask[1], dqmag_dF, &factor);
+      factor *= (1. - ratio);
+      factor += ratio;
+      dq_gradp *= factor;
+      dq_dH *= factor;
+      pre_delP *= factor;
+      vis_w /= factor;
+    }
```
wortiz, line 4084: This is the remaining exodus errors from the current state. Needs a guard/if

wortiz:
Were you having difficulties with the normals not being outward facing? It does seem like we set `T` to provide that in the regular non-shell elements. I'm not sure why the change to a constant T for normal shells.

rbsecor:
Some of these mods I had forgotten about.  At one time, I was trying to do edge normals from the shell element instead of the bulk element but later gave up on that.  I'll do some code cleanup to remove some unnecessary stuff.

rbsecor:
OK, I think I'm finished with a little clean-up.  Hopefully this is ready to merge now.

PR #421, opened 2023-02-15
Bugfixes for strain rate in belly_flop

rbsecor:
Added post-processing options for solid rate-of-strain invariants.  These turned out to be always zero which was traced down to the use of fv_dot->grad_d.  This is apparently never set, leading strain rate variables to be always zero.  This and others were replaced by fv->grad_d_dot and similar replacements.

Added some user routines for solid viscosity functions.  Current models have WLF style viscosity vs. temperature behavior.


src/mm_fill_terms.c
```diff
@@ -13740,6 +13740,17 @@ double density(DENSITY_DEPENDENCE_STRUCT *d_rho, double time)
...
     }
 
+  } else if (mp->DensityModel == DENSITY_THERMEXP) {
+    rho = mp->u_density[0] / (1. + mp->u_density[1] * (fv->T - mp->reference[TEMPERATURE]));
+    if (d_rho != NULL) {
+      var = TEMPERATURE;
+      for (j = 0; j < ei[pg->imtrx]->dof[var]; j++) {
+        d_rho->T[j] = rho * mp->u_density[1] /
+                      (1. + mp->u_density[1] * (fv->T - mp->reference[TEMPERATURE])) *
+                      bf[var]->phi[j];
```
wortiz, line 13750: I think this is missing a minus sign, unless I'm misreading something.

rbsecor: That seems right.

rbsecor:
Since belly_flop is called "all the time", I'm thinking I should only do the "dot" calculations for transient problems.  i.e., don't slow down steady calculations doing strain-dot calculations.  Good idea?

wortiz:
> Since belly_flop is called "all the time", I'm thinking I should only do the "dot" calculations for transient problems. i.e., don't slow down steady calculations doing strain-dot calculations. Good idea?

That would probably be better as we call these inside the Gauss point loop

I'm thinking these d_grad_d_dot terms need to be multiplied by `(1 + 2 * tran->theta) / tran->delta_t ` to account for the esp_dot->d:

https://github.com/rbsecor/goma/blob/210d32c9c06e97a21e11eb154bc359ac9ac4bb18/src/mm_fill_solid.c#L202-L208

I'm not sure if we have good Jacobian checking for transient problems or not.

rbsecor:
It looks like you're correct.  Seems it's needed for the Newmark beta
scheme which I think is only used for DYNAMIC_LAGRANGIAN.  Do we have a
test suite problem for DYNAMIC_LAGRANGIAN?  Anyway, let me fix this up
and we can maybe merge next week.

Robert

wortiz:
> I think is only used for DYNAMIC_LAGRANGIAN.  Do we have a
test suite problem for DYNAMIC_LAGRANGIAN?

We have a few it looks like.

I think what I had commented on earlier I had missed the ARBITRARY if check (I guess d_phi should've been a hint), I guess it probably isn't used there as we only do LINEAR/NONLINEAR 

The `tran->theta` term should be for our CN vs BE time integration, I'm not sure if Newmark-beta works or what we have examples for.

It looks like for your Kelvin-Voigt model the load_fv_mesh_derivs would be the appropriate place but it is blocked by `discontinuous_porous_media` check so I guess you would also have to enable for your models as well to get the right Jacobian entries.

https://github.com/rbsecor/goma/blob/210d32c9c06e97a21e11eb154bc359ac9ac4bb18/src/mm_fill_terms.c#L13169-L13194

```
        /*N.B. PRS: need to actually furbish this for Newmark Beta schemes */
        if (transient_run && discontinuous_porous_media) {
          /* fv->d_grad_d_dot_dmesh[p][q] [b][j] += (1.+2.*tran->theta)*bfv->grad_phi_e[j][b]
           * [p][q]/tran->delta_t;*/

          fv->d_grad_d_dot_dmesh[0][0][b][j] +=
              (1. + 2. * tran->theta) * bfv->grad_phi_e[j][b][0][0] / tran->delta_t;
```

wortiz:
.runguts

wortiz:
The exodus error was a negligible change in CONVORDER, I changed the tolerance for CONVORDER in the test suite to accommodate.

rbsecor:
Since belly_flop is called "all the time", I'm thinking I should only do the "dot" calculations for transient problems.  i.e., don't slow down steady calculations doing strain-dot calculations.  Good idea?

rbsecor:
If this version gets a guts OK, I would be ready to merge if you are.  I had to fix some issues I was having with lame_mu Tables and USER thermal_expansion.

wortiz (approved):

PR #434, opened 2023-07-17
USNCCM17 version

rbsecor:
Updates for solid thermal expansion, orthotropic expansion, Standard linear solid (SLS) model, etc.  First check-in of some moving boundary shear-thinning lubrication routines -- these have not been tested yet.  The thermal expansion adds outfit Goma for thermal expansion that is temperature dependent, i.e., expansion ~ exp(Int(alpha*dT)) -1.  Corresponding mods are made to ORTHOTROPIC expansion which is currently transversely isotropic, i.e. one direction is different from the isotropic value.  Ultimately a second direction would be needed for true orthotropic expansion.  The SLS model has seen limited testing -- there may still be Jacobian mishaps in HOOKEAN, KELVIN-VOIGT base models.


wortiz:
.runguts

src/mm_post_proc.c
```diff
@@ -1784,14 +1787,22 @@ static int calc_standard_fields(double **post_proc_vect,
...
         }
       }
+      if (DOUBLE_ZERO(q_mag)) {
+        q_mag = 1.;
+      }
       if (fv->x0[2] <= radius) {
         dir = 0;
       } else {
-        // double tmp = modf((fv->x0[2] - radius) / (2 * radius - overlap), &sign);
+        double tmp;
```
wortiz, line 1796: This appears to be the last warning:

```
variable ‘tmp’ set but not used [-Wunused-but-set-variable]
```

src/mm_fill_terms.c
```diff
@@ -295,10 +307,51 @@ int assemble_mesh(double time,
...
+    if (af->Assemble_Jacobian) {
+      memset(dTT_dx, 0, sizeof(double) * DIM * DIM * DIM * MDE);
+      for (a = 0; a < VIM; a++) {
+        for (b = 0; b < VIM; b++) {
+          for (mode = 0; mode < vn->modes; mode++) {
+            for (p = 0; p < dim; p++) {
+              var = MESH_DISPLACEMENT1 + p;
+              if (pd->v[pg->imtrx][var]) {
+                for (j = 0; j < ei[pg->imtrx]->dof[var]; j++) {
+                  dTT_dx[a][b][p][j] += fv->grad_S[mode][p][a][b] * bf[var]->phi[j];
```
wortiz, line 337: Assuming this corresponds to:

```
TT[a][b] += fv->S[mode][a][b];
```

Should this not just be 0 as in there is no displacement contribution.

Maybe I'm missing something else.

rbsecor: Hmm...  I think that's correct.  I'll update things.

wortiz (approved):

PR #439, opened 2023-10-19
Lubrication Wall Effect

rbsecor:
Check-in of lubrication wall effects - first iteration.  A Newtonian style increased liquid permeability near the wall which is implemented via a modified lubrication height.  Power-law and moving wall implementations yet to come.
Also:
1 - Optional output variable linear regression over the whole domain
2 - Parallel fixes for TORQUE in mm_flux.c
3 - Numerous shear-thinning lubrication LS Jacobian fixes


wortiz:
.runguts

wortiz:
.runguts

rbsecor:
Based on local guts results, it seems there's some Jacobian issues with some porous shell problems which I'll take a look at.

wortiz:
.runguts

PR #441, opened 2023-11-17
revert change to CONSTANT thermal expansion to preserve backwards compat.

wortiz:
@rbsecor this creates a new model CONSTANT_DV with your changes

Users were using old behavior for their models. Maybe we can remember to switch constant behavior on a Goma 8 release but reverting to keep backward compatibility for now.


wortiz:
Merged for now we can rename CONSTANT_DV in a future commit if needed

PR #449, opened 2023-12-13
Lub Curvature Jacobian fixes

rbsecor:
Corrections to Lubrication Jacobian entries, principally w.r.t. Curvature and Level Set field.

- Lub_Curv_Diff added as a material input to control this smoothing function
- ...


rbsecor:
I might suggest some re-baselining with this one.  It' s possible I fixed some minor Jacobian errors, resulting in slightly different convergence rates -- but some changes of sign with one test problem.

wortiz:
I think two of them are just CONVRATE CONVORDER and can probably be rebaselined. 

`p_lub_pore` appears to have spurious currents with the PR we can look at it during the Goma meeting this AM.

wortiz:
Actually your commit looks like it corrects some spurious currents so maybe it would be better rebaselined.

src/mm_fill_shell.c
```diff
@@ -13936,18 +13940,22 @@ int assemble_lubrication_curvature(double time,            /* present time value
...
+        }
       }
       mass *= det_J * wt * h3 * pd->etm[pg->imtrx][eqn][(LOG2_MASS)];
 
       /* Assemble diffusion terms */
       diff = 0.0;
       if (pd->e[pg->imtrx][eqn] & T_DIFFUSION) {
         for (a = 0; a < VIM; a++) {
-          diff += *hsquared * gradII_kappa[a] * grad_II_phi_i[a];
+          diff += hsquared[a] * gradII_kappa[a] * grad_II_phi_i[a];
```
wortiz, line 13955: It looks like this is causing the main difference, but the fix you've done is actually correct so maybe that is why we've seen improvements. I've rebaselined the test suite to this.

wortiz:
.runguts

PR #450, opened 2024-01-16
White metzner

wortiz:
Bring back White-Metzner models through non-constant Polymer Time Constant cards.

Currently supports CONSTANT, POWER_LAW, CARREAU


wortiz:
.runguts

wortiz:
.runguts

PR #455, opened 2024-03-28
Shell curvature updates

rbsecor:
Added a number of shell curvature options:
 - curvature field advection with Lubrication Curvature Relaxation
 -  option for Petrov-Galerkin on curvature equation with Lubrication Curvature Weight Function
 -  Mass lumping or not with Lubrication Curvature MassLump
 -  option for ls->delta*normal vs. grad_II H in surface tension term


wortiz:
If you're having clang-format issues the tester is using clang-format-17

You can download a binary here:
https://github.com/llvm/llvm-project/releases/tag/llvmorg-17.0.6

The x86_64-linux one and it will have it under the bin folder or you can use apt to install it https://apt.llvm.org/

rbsecor:
Thanks!  That was very helpful.

src/mm_shell_util.c
```diff
@@ -3493,7 +3493,7 @@ void calculate_lub_q_v(const int EQN, double time, double dt, double xi[DIM], co
   int i, j, k, jk, w;
   dbl H;
   dbl veloL[DIM], veloU[DIM];
-  dbl mu, dmu_dc = 0., srate = 0.;
+  dbl mu, dmu_dc = 0., dmu_dT = 0., srate = 0.;
```
wortiz, line 3496: Compiler warning dmu_dT set but not used.

wortiz (changes requested):

wortiz (approved):

PR #468, opened 2024-07-09
Lubrication Logarithm Interpolation

rbsecor:
Added Logarithmic interpolation across the level-set interface for shell lubrication equations.  This seems to make the viscosity behavior for shear-thinning liquids much smoother across the LS interface.

It's invoked via
Lubrication Level-Set Interpolation = LOGARITHMIC
i.e., interpolated value pp = p0^1-H(F) * p1^H(F)

The default is the standard linear interpolation, pp = p0 * (1-H(F)) + p1 * H(F)

At present, only the lubrication flow q(gradP) can be interpolated this way.


rbsecor:
I probably need to add a test suite problem for LS Logarithmic interpolation... will work on it.

src/mm_viscosity.c
```diff
@@ -3234,6 +3234,8 @@ int ls_modulate_viscosity(double *mu1,
     factor += ratio;
   } else if (model == TIME_RAMP) {
     factor *= (1. - ratio);
+  } else if (model == TIME_RAMP) {
```
wortiz, line 3237: Seems like this accidentally got added twice

wortiz (changes requested):

wortiz (approved):

wortiz:
.runguts

PR #470, opened 2024-07-29
Lub log ls2

rbsecor:


wortiz:
.runguts

wortiz:
.runguts

wortiz:
.runguts

rbsecor:
FYI, The viscous dissipation terms are now in the shell energy equation, so the lub_energy problem will need to be re-baselined at some point.  But, there appear to be some missing Jacobian terms for the viscous dissipation terms yet so probably wait until I get those resolved.

wortiz:
I didn't see anything that stuck out. Do you still think you have some missing Jacobian terms?

rbsecor:
The Jacobian terms are in, but I just had a thought that I should connect them up to the Heat Source card so we could turn them on or off from the input file.  I'll let you know when Iḿ done with that. - Thanks

rbsecor:
OK, the Lubrication viscous dissipation terms are connected to the *.mat files.  We would need a
Heat Source    =  VISC_DISS  1.0
in the *.mat files to get the same answers as last time around.

wortiz:
.runguts

wortiz:
.runguts

rbsecor:
Outfitted HEAT_ENERGY volumetric integration for shells, and added HEAT_ENERGY_POS and HEAT_ENERGY_NEG.  And tried to simplify some of the logic.

BTW, I'm thinking the guts test suite does not compare integrated fluxes or volumetric integration, which seems OK since we're comparing the exoII file output.  But this does mean we would not be testing whether the integrated fluxes or volumetric integration routines have changed...  I.e., whether these changes provide any different results.  It doesn't seem to hard to add a volumetric or flux integration check if it's not there -- just add the flux/volume results to the files we're diffing, perhaps.

wortiz:
I have a couple of custom tests that do the flux out and such but they have to be added explicitly. I'll add to lub_energy

wortiz:
Guts on my machine was giving segfaults I found two address sanitizer errors that are fixed with my commit. They actually came from the lub_CYmold problem oddly enough.

I also added the flux/volume tests so there should be another custom test when this is tested.

wortiz:
I think my commit https://github.com/goma/goma/pull/470/commits/2af8939b394619e9dae940abb0e549f48aafa7ad should just affect your LS_CURV_NOBC which previously had a memory bug. 

Let me know if this looks ready to merge.

PR #484, opened 2024-09-19
Disable mesh sensitivities for SA model until they can be fixed.

wortiz:
Disable


wortiz:
This should be the last bugfix backport for training.

PR #485, opened 2024-09-24
Kelvin-Voigt solid model fix

rbsecor:
Fix for the Kelvin-Voigt rate-of-volume-strain term, which showed slow convergence.  The fix was applied only to the 3D model for now.  Plane Strain and plane stress models will probably need to be updated.


rbsecor:
I believe these Kelvin-Voigt fixes are ready for review.  I.e., Roxana's problem seems to work fine in 2D and 3D as long as I initialize with a neo-Hookean solution.  Initializing with zero displacements seems to cause convergence issues.  I still need to create a test suite problem for Kelvin-Voigt.  I'm thinking of something more basic than Roxana's drop -- maybe a twisting cylinder so that we sample shear deformations and not volume compression/expansion.

wortiz:
I think this looks ok to merge. Do you want me to wait til you have a test suite problem?

wortiz (approved):

PR #488, opened 2024-10-31
Making Aztec/Amesos/Epetra optional for upcoming releases of Trilinos

wortiz:
Probably needs some more testing...


wortiz:
.runguts

PR #490, opened 2024-11-13
SHELL_CONC_LS_BC

rbsecor:
Added a SHELL_CONC_LS_BC which is a STRONG_INT, CROSS_PHASE bc for communicating a species concentration level to a material based on the position of the LS interface in the adjacent material.  Also, some lubrication Heaviside function shifting fixes.


rbsecor:
Most of the test suite failures looked minor, mostly slight differences in the lubrication problems which could attributed to the change in the curvature field modulation, i.e. the diffusion of curvature is now only active in the level-set interface vicinity.  These problems could probably be re-baselined.  But I do see the lub_CYmold problem had a different number of time steps -- that one probably requires more looking into...

wortiz:
>  But I do see the lub_CYmold problem had a different number of time steps -- that one probably requires more looking into...

I think this might be the narrow region and the time step control causing reduced time step size.

I don't think I saw anything else that would cause this.

wortiz (changes requested):
The changes I mention in the code fix the rest of the problems.

For lub_CYmold:

So it looks like the diffusion term is probably causing issues with the timestepping.

It looks pretty bad outside of the interface, for diffusion would we rather want the opposite of curvX (low near interface and high outside of interface)? 

Picture of issues near edges of curvX

image

wortiz:
To clarify the image above was with the relaxation turned to 0 as I was trying to figure out where the change was coming from.  Overall the diffusion scaling by curvX doesn't really seam to change inside the region of lsi->alpha but does seem to cause some issues outside of it which seems to be causing the time step differences.

rbsecor:
The part with the time step control on lub_CYmold seems to be due to the curvature field being near zero.  If I compile with option_definition(VAR_UPDATE_UNITY_SCALE ON), this problem is back to 10 timesteps instead of 16.  But, of course, there are a bunch of minor exoII diff failures due to different time step control.  Maybe I'll think a bit about different ways to enforce some appropriate scaling.

wortiz:
If I limit the time_step_control to the level set region of interest I still end up with more steps but it does seem to reduce the error:

With limited region:
```
YUK 8.0e-04 > 1.5 5.0e-04 [, 3.4e-04, 7.3e-04]
Predictor-corrector norm: [,   1 F  , 1 SHELL]
```

Without limited region:
```
YUK 1.1e-03 > 1.5 5.0e-04 [, 2.9e-04, 1.0e-03]
Predictor-corrector norm: [,   1 F  , 1 SHELL]
```

If I turn off the time_step_control for SHELL_LUB_CURV:

```
OK  3.6e-04 < 1.5 5.0e-04 [, 3.6e-04, 3.9e-06]
Predictor-corrector norm: [,   1 F  , 1 SHELL]
```

I added this to time step control around line 603:

```
    if ((eqn == SHELL_LUB_CURV || eqn == SHELL_CURVATURE2) &&
        upd->matrix_index[FILL] == pg->imtrx) {
      int ji = Index_Solution(inode, FILL, 0, 0, vd->MatID, upd->matrix_index[FILL]);
      valid = FALSE;
      if (ji != -1 && x[ji] < ls->Length_Scale * 0.5) {
        valid = TRUE;
      }
    }
```

I do still end up with 1 extra time step turning off time_step_control for SHELL_LUB_CURV so it seems to not just be the projection equation causing some of the time step restriction.

rbsecor:
After spending some time looking at the error norms for the time-step control and not finding much (i.e., it seems the highest time-step errors are almost always in the FILL equation), I tried changing the time integration method from trapezoidal rule to Backward Euler in the lub_CYmold test problem.  This seemed to eliminate any failed time steps.  So, this seems like one option -- but would require re-baselining the problem since the time steps would be slightly different.

wortiz:
I guess I have two questions:

1. Are you seeing better behavior in other problems with this curvX scaling on the diffusion term for the curvature?
2. Do you want this to be the only option for your `Lub_Curv_Modulation` or do you want to have an option for the old behavior.

Right now this only affects your problem so if you see better results with the scaling on we can rebaseline on backward Euler or do something like increase time step error. 

It looks like the previous version was failing a time step as well but that was on FILL like you had mentioned with your tests.

I'll leave it up to you if you want me to rebaseline your problem or if you want to add a mp->Lub_Curv_Diffusion_Modulation or something else to preserve old behavior.

I think you've preserved old behavior on other problems with the Jacobian fixes.

wortiz:
If you are seeing better behavior and are unlikely to use the separate tag we can forgo that and instead rebaseline just let me know. I only messed around with the small test case so I wasn't sure if it was improving or not.

One thing I did see in the code is that the SUPG term was only added to the advection term which can cause some extra diffusion. In the image below a small advection equation moving a square wave is shown. Yellow is SUPG only on the advection term while black is on both mass and advection term (no diffusion and source, and purple is with an additional shock capturing term). So you could add the SUPG term to the mass term in the future if you still see excess diffusion. Assuming you are using that Pseudo-Relaxation card

image

rbsecor:
Umm, so far in the places I've looked in Goma it seems the SUPG weighting is only used on the advection term (most of the shell equations and assemble_energy, for example).  Is this perhaps a wider Goma issue?  I assume changing it everywhere would have a substantial effect on the test suite.  I guess we should do some testing.

I did also notice that I should make SHELL_LUB_CURV_2 look the same as SHELL_LUB_CURV, which I don't think it currently does...

rbsecor:
I attempted to update the SUPG option for lubrication equations (mostly SHELL_LUB_CURV and SHELL_ENERGY).  At this point there seems to be a parallel issue with lub_CYmold -- runs OK with 1 processor, time_step_error is different for multiple processors.

include/mm_mp_structs.h
```diff
@@ -775,6 +775,7 @@ struct Material_Properties {
   int Lub_Curv_Modulation;
   int Lub_LS_Interpolation;
   int Lub_Curv_Combine;
+  int Lub_Isotropic_Curv_Diffusion;
```
wortiz, line 778: The parallel bug might be this missing from dp_vif.c

rbsecor:
It seems the SHELL_LUB_CURV equation has more equation multipliers now, so breaking of test suite problems should be expected.  Oddly I can get close to the pshell_open solution by turning off DIVERGENCE term, i.e., no curvature generation.  How did this problem ever work?  I also realized the SHELL_LUB_CURV LOG2_BOUNDARY multiplier was never read.  No wonder the curvature bcs seemed to have no effect.  Probably some reworking of the shell curvature test suite problems is in order -- at a minimum changing the equation multipliers.

wortiz:
> Oddly I can get close to the pshell_open solution by turning off DIVERGENCE term, i.e., no curvature generation. How did this problem ever work? 

It is possible I messed this problem up during one of our rebaselines it might be that it was working before.

pshell_open and pshell_closed have had a lot of issues over the years so maybe it got broken along the way.

> I also realized the SHELL_LUB_CURV LOG2_BOUNDARY multiplier was never read. No wonder the curvature bcs seemed to have no effect. Probably some reworking of the shell curvature test suite problems is in order -- at a minimum changing the equation multipliers.

Does your ETM defaults work okay here when none are set or should I look at setting them manually?

I feel like the boundary ETM has only ever bitten us. I think not every BC checks for that multiplier but it feels like the only time we would ever want to use that multiplier is when checking Jacobians but it feels like that could be accomplished as well with just removing BCs. Unless I'm missing another use case for this ETM.

rbsecor:
This would seem to be a problem for LOG2_BOUNDARY in bc_integ.c
                /*
                 *   Add in the multiplicative constant for corresponding to
                 *   all boundary conditions, except for certain special
                 *   cases
                 */
                if (bc_desc->method == WEAK_INT_SURF && bc->BC_Name != PSPG_BC &&
                    bc->BC_Name != VELO_SLIP_SOLID_BC && bc->BC_Name != ELEC_TRACTION_BC &&
                    bc->BC_Name != QSIDE_LS_BC && bc->BC_Name != LS_ADC_BC &&
                    bc->BC_Name != SHEAR_TO_SHELL_BC && bc->BC_Name != POR_LIQ_FLUX_FILL_BC &&
                    bc->BC_Name != DARCY_LUB_BC && bc->BC_Name != SHELL_TFMP_FREE_LIQ_BC &&
                    bc->BC_Name != SHELL_TFMP_NUM_DIFF_BC && bc->BC_Name != SH_SDET_BC &&
                    bc->BC_Name != SH_MESH2_WEAK_BC &&
                    bc->BC_Name != SHELL_LUBRICATION_OUTFLOW_BC) {
                  weight *= pd->etm[pg->imtrx][eqn][(LOG2_BOUNDARY)];
                }
i.e., hits most bcs and I'm assuming the etm will be zero if it's not read...

I don't think I added ETM defaults to the SHELL_LUB_CURV yet.  I'll take a look and probably add some defaults.

wortiz:
> i.e., hits most bcs and I'm assuming the etm will be zero if it's not read...

Yeah it looks like it hits everything except for strong BCs and special BCs. I just meant the boundary etm seems like a _footgun_ if you forget to add it when you put in an equation and then you need a weak BC later. 

I was just thinking that it might need to be removed in the future.

rbsecor:
Added SHELL_LUB_CURV etm defaults and corrected a property default mis-step.

rbsecor:
There seems to be definitely something amiss in the pshell problems -- lots of apparent Jacobian errors in lubrication terms...  I'll see what I can find.

wortiz:
Something funny happened on lub_CYmold with the default ETMs being turned on. The problem now takes 15 hours because its using all the timesteps, I guess I should remove minimum resolved timestep from that one.

Convergence looks like maybe a bad Jacobian is contributing, it made 13 of 4999 timesteps.

rbsecor:
OK, I've duplicated this behavior for lub_CYmold.  I'll investigate.

wortiz:
Seems like quite a few problems had 0 for the diffusion.

Were you getting the same number of timesteps with lub_CYmold as before this patch without the BCs? Wondering if I missed something I was still getting an extra timestep.

I thought maybe it was the penalty portion of the BC causing issues but that didn't really change the behavior turning that off.  I've disabled the BCs on the latest guts version for now.

rbsecor:
I was getting 10 timesteps for lub_CYmold.  In looking through, I don't see any parameter changes -- just removed the etms from the Curvature eqn so we were getting the defaults.  I did fix the reference temp aprepro error but that shouldn't have any influence.

rbsecor:
If would appear that the extraneous curvature diffusion boundary term had the wrong sign.  It seems like the LUB_CURV_NOBC bcs now work (or rather don't cause catastrophic errors).

wortiz:
.runguts

wortiz:
I've turned them back on for lub_CYmold.

wortiz (approved):

PR #494, opened 2025-01-16
Gas egress

rbsecor:
Adding "gas egress" channels as a means to promote wetting (lower dynamic contact angles) with multi-phase lubrication calculations where sidewall drag is included.  On the liquid side of the LS interface, the lubrication gap is reduced to capture the increased drag of the sidewall.  On the gas side, the opposite is true.  The gap near the side wall is increased to promote gas egress from the interface zone and thereby lower the apparent dynamic contact angle.


rbsecor:
I forgot about this one for a while.  It seems I need to look through and find why lub_CYmold is behaving differently.

wortiz:
> I forgot about this one for a while. It seems I need to look through and find why lub_CYmold is behaving differently.

I took a look and it seems the changes in two files are causing the differences: mm_fill_shell.c and mm_shell_util.c

The differences look pretty minor and are mostly in the velocity post processing near the interface.

I think the mm_shell_util.c probably looks ok and can explain some of it just by the switch to lsi->near.  I didn't try messing around to see what exactly was causing the changes.

rbsecor:
Ah yes, there was a slight bug in mm_shell _util (and also mm_std_models_shell as I recall) where I had used the LS alpha as equal to the LS Length Scale instead of alpha= 0.5*LSLS.  And I'm 99% sure the sign changes in the curvature eqn in mm_fill_shell are correct as we're adding a curvature term that integration-by-parts has been applied to another curvature term where it has not.  Of course, I don´t know that this would be captured in the test suite problems.

I've been trying to compare the lub_CYmold results I get on my machine to those from Jenkins, and oddly, there are differing number of timesteps.

rbsecor:
It looks like the aprepro warning messages in lub_CYmold can be eliminated by using these "apprepro-ed" mat files.  But, it seems these warnings are irrelevant to the problem execution...
[carbo_tube_LS_app.mat.txt](https://github.com/user-attachments/files/18811940/carbo_tube_LS_app.mat.txt)
[carbo_LS_app.mat.txt](https://github.com/user-attachments/files/18811941/carbo_LS_app.mat.txt)

wortiz:
.runguts

PR #497, opened 2025-02-19
Add logarithmic interpolation of properties across level set interface

tjiptowi:


wortiz:
This doesn't seem to have errors anymore after rebaselining for #494 I haven't found any reason why this would've changed lub_CYmold

wortiz (approved):

PR #507, opened 2025-03-21
update for clang-format-18

wortiz:


wortiz:
Newer versions of clang format have changed the AlignTrailingComments behavior so we'll just update to match.

Clang 18 has been out for a while now.  Similar behavior exists in Clang 19. So update so we don't get stuck on an old version.

PR #509, opened 2025-03-28
Improved facet based reinitialization of level sets

wortiz:
Previously Huygens based reinitialization of the level set seemed to have some odd behaviors such as large mass loss especially in 3D.

This takes a marching squares and marching cubes approach to finding facets in 2D and 3D and then finds nearest neighbors using facet centroids to compute distances to each reinitialized node.

It seems reasonably fast especially in parallel and should improve reinitialization in 3D:

TODO:

- [x] Tetrahedron case
- [x] Fix print statements
- [x] Better testing
- [x] Cleanup


wortiz:
.runguts

PR #511, opened 2025-04-11
Native MUMPS interface

wortiz:
Start of a native mumps interface for Goma.

- [ ] Add to other solver calls
- [ ] Check that we handle failure
- [ ] Add input options for mumps
- [ ] update documentation

Seems faster than either Amesos choice at least.


wortiz:
.runguts

wortiz:
.runguts

wortiz:
.runguts

wortiz:
.runguts

PR #528, opened 2025-07-14
add DOUBLE_FILLET boundary condition

wortiz:
This is a modification of the DOUBLE_RAD BC, keeping the old one for backwards compatibility.

Modification originally by @ChanceCP.

Diagram from @trkenne


wortiz:
.runguts

PR #529, opened 2025-07-18
Pixel to mesh fixes

wortiz:
- Supports parallel
- Bug fix on Jacobian not being calculated correctly
- Bilinear interpolation fixed
- image_to_vox.py in scripts directory

Probably still needs some changes just making sure it doesn't break stuff


wortiz:
.runguts

PR #550, opened 2025-10-07
TPL script: Update OpenBLAS to use CMake and print time taken

wortiz:
Also some fixes for redhat like systems


wortiz:
.runguts

PR #564, opened 2026-05-06
Some TPL updates

wortiz:
Last Trilinos with Amesos+Aztec, with a patch for MacOS and newer compilers if it comes up later.

New MUMPS


wortiz:
.runguts

PR #566, opened 2026-05-22
Film casting model 2.5D

wortiz:
Adds a film height equation with new momentum assemblies and support for generalized Newtonian and viscoelastic flows for film casting flows.

Needs some cleanup but checking if anything is broken.


wortiz:
.runguts

